Issue #7, opened 2018-06-22
Cross-build Scala versions


hugo-vrijswijk:
To ensure the project works with multiple Scala versions, we should cross-build (and test) against them.

Based on the [recent Jetbrains survey](https://www.jetbrains.com/research/devecosystem-2018/scala/) I think we should support at least Scala 2.11 and 2.12. And maybe when there is time look into what would be needed for 2.13.

PR #8, opened 2018-06-22
Add .travis.yml file

### LINKED ISSUE #1: Automated build
We need a build server so we compile/test and eventually publish artifacts. 

I think travis.ci is the way to go because it's public ans widely used and supported.

We also need to make sure the build succeeds on multiple versions of scala so we need to enable cross compiling in the build.sbt

Would be nice if we could add a build status badge as well.

hugo-vrijswijk:
Fixes #1.

Runs a commit with scala 2.12.6. The [default test script for travis sbt projects](https://docs.travis-ci.com/user/languages/scala/#sbt-Default-Script-Command) is `sbt test`, so no further script commands are needed as of yet

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,19 @@
+# Use container-based infrastructure
+sudo: false
+
 language: scala
+
+# These directories are cached to S3 at the end of the build
+cache:
+  directories:
+    - $HOME/.ivy2/cache
+    - $HOME/.sbt
 
 scala:
    - 2.12.6
+   - 2.11.12
+
+before_cache:
+  # Cleanup the cached directories to avoid unnecessary cache updates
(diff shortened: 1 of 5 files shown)


legopiraat:
Let's add 2.11.12 as well to make it works on both versions. 
We could easily fix #7 in this merge request as well

hugo-vrijswijk:
I've taken a look at the [SBT Travis manual ](https://www.scala-sbt.org/1.0/docs/Travis-CI-with-sbt.html)and added both Scala 2.11.12 and dependency caching (which should hopefully speed up the build a bit)

hugo-vrijswijk:
Added the cross-compilation to the SBT build settings (and fixed a compile error for 2.11)

PR #18, opened 2018-06-24
Add a way to exclude files from mutating

legopiraat:
Added a config entry to exclude files from beeing mutated

### DIFF
--- a/core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
+++ b/core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
@@ -13,7 +13,7 @@ class FileCollector(implicit config: Config) extends SourceCollector {
   private[this] val toExcludeFiles: Seq[File] = toFileList(config.excludedFiles)
 
   override def collectFiles(): Iterable[File] = {
-  toMutateFiles.filterNot(file => toExcludeFiles.contains(file))
+    toMutateFiles.filterNot(file => toExcludeFiles.contains(file))
   }
 
   private[this] def toFileList(files: Seq[String]): Seq[File] = {
--- a/core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
+++ b/core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
@@ -9,9 +9,14 @@ trait SourceCollector {
 
 class FileCollector(implicit config: Config) extends SourceCollector {
 
-  override def collectFiles(): Iterable[File] =
-    config.baseDir
-      .glob(config.files.mkString("{", ",", "}"))
-      .toIterable
+  private[this] val toMutateFiles: Seq[File] = toFileList(config.files)
+  private[this] val toExcludeFiles: Seq[File] = toFileList(config.excludedFiles)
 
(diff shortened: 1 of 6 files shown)


hugo-vrijswijk:
Looks cool! But maybe we should exclude files by the pattern having a `!` at the start, similar to [Stryker](https://github.com/stryker-mutator/stryker/blob/master/packages/stryker/README.md#configuration). And add documentation for it in the [configuration](https://github.com/stryker-mutator/stryker4s/blob/master/docs/CONFIGURATION.md)

core/src/test/scala/stryker4s/mutants/findmutants/FileCollectorTest.scala
```diff
@@ -66,9 +67,61 @@ class FileCollectorTest extends Stryker4sSuite {
...
+        implicit val config: Config = Config(
+          files = Seq("**/someFile.scala", "**/secondFile.scala"),
+          excludedFiles = Seq("**/*.scala"),
+          baseDir = filledDirPath)
+
+        val sut = new FileCollector()
+
+        val results = sut.collectFiles()
+
+        results should have size 0
```
hugo-vrijswijk: You can also use `should be (empty)`, which might be a little cleaner

legopiraat:
i've added the documentation :)

For now i would prefere to keep it without the `!` keeps it simpler and might be a improvement for the future.

core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
```diff
@@ -9,9 +9,14 @@ trait SourceCollector {
...
-      .toIterable
+  private[this] val toMutateFiles: Seq[File] = toFileList(config.files)
+  private[this] val toExcludeFiles: Seq[File] = toFileList(config.excludedFiles)
 
+  override def collectFiles(): Iterable[File] = {
+    toMutateFiles.filterNot(file => toExcludeFiles.contains(file))
+  }
+
+  private[this] def toFileList(files: Seq[String]): Seq[File] = {
+    files.flatMap(config.baseDir.glob(_))
```
hugo-vrijswijk: What happens here if multiple patterns match the same file? Wouldn't it be included multiple times?

nicojs:
Using `!` is more expressive than having a separate `exclude` setting. You could do something like this:

```js
files: ['src', '!src/generated', 'src/generated/myFile.scala']
```

(don't worry about the syntax not being valid sbt ;))

legopiraat:
If you both think it should be with a `!` then i guess ill refactor it.

core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
legopiraat: Think this might of already been a problem. I've added the `distinct` to make sure this doesn't happen.

core/src/test/scala/stryker4s/mutants/findmutants/FileCollectorTest.scala
legopiraat: done

hugo-vrijswijk:
All looks good to me!

PR #22, opened 2018-06-30
Mutation score decimal rounding

legopiraat:
Fixes: #22 

Refactored the mutation score calculation so it will be calculated outside the reporter (which should be done outside the reporter in my opinion anyway). It is now formatted based on 2 decimals. 

The commits are a bit akward but we should squash anyway.

### DIFF
--- a/core/src/test/scala/stryker4s/run/process/ProcessMutantRunnerTest.scala
+++ b/core/src/test/scala/stryker4s/run/process/ProcessMutantRunnerTest.scala
@@ -87,7 +87,7 @@ class ProcessMutantRunnerTest extends Stryker4sSuite {
       )
     }
 
-    it("should return a mutationScore of 66.66 when 2 of 3 mutants are killed") {
+    it("should return a mutationScore of 66.67 when 2 of 3 mutants are killed") {
       val testProcessRunner = new TestProcessRunner(Success(1), Success(1), Success(0))
       val sut = new ProcessMutantRunner(testProcessRunner)
       val mutant = Mutant(1, q"0", q"zero")
--- a/core/src/test/scala/stryker4s/run/process/ProcessMutantRunnerTest.scala
+++ b/core/src/test/scala/stryker4s/run/process/ProcessMutantRunnerTest.scala
@@ -20,7 +20,7 @@ class ProcessMutantRunnerTest extends Stryker4sSuite {
     it("should return a Survived mutant on an exitcode 0 process") {
       val testProcessRunner = new TestProcessRunner(Success(0))
       val sut = new ProcessMutantRunner(testProcessRunner)
-      val mutant = Mutant(0, q"4", q"5")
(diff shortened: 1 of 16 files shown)


core/src/test/scala/stryker4s/run/process/ProcessMutantRunnerTest.scala
```diff
@@ -76,10 +79,34 @@ class ProcessMutantRunnerTest extends Stryker4sSuite {
...
       testProcessRunner.timesCalled.next() should equal(2)
+
+      result.mutationScore shouldBe 100.00
       result.results should contain only(
         Killed(1, mutant, Paths.get("simpleFile.scala")),
         Killed(1, secondMutant, Paths.get("simpleFile.scala"))
       )
     }
+
+    it("should return a mutationScore of 66.66 when 2 of 3 mutants are killed") {
```
hugo-vrijswijk: Shouldn't this be 66.67?

core/src/test/scala/stryker4s/stubs/TestProcessRunner.scala
```diff
@@ -5,11 +5,11 @@ import stryker4s.run.process.ProcessRunner
...
 import scala.util.Try
 
-class TestProcessRunner(returns: Try[Int]) extends ProcessRunner {
+class TestProcessRunner(returns: Try[Int]*) extends ProcessRunner {
   val timesCalled: Iterator[Int] = Iterator.from(0)
 
   override def apply(command: String, workingDir: File, envVar: (String, String)): Try[Int] = {
     timesCalled.next()
-    returns
+    returns(envVar._2.toInt - 1)
```
hugo-vrijswijk: What does this do? Does it return the envVar as the result for the exit code? Maybe a comment would be nice

core/src/test/scala/stryker4s/run/process/ProcessMutantRunnerTest.scala
legopiraat: True it was late my maths were off.

core/src/test/scala/stryker4s/stubs/TestProcessRunner.scala
legopiraat: Added some comments on how this works.

What do you think about starting the test run numbers on 0 instead of 1? This would allow us to remove the -1 in the test run.

hugo-vrijswijk: Hmm, maybe. In my mind mutation run 0 was always "reserved" for no mutations. But I don't see why not. What do StrykerJS and Stryker.NET do? I think it's good to have it the same as them

nicojs: Ah, I see. Couldn't we just use `ACTIVE_MUTANT=''` to clear the env variable? Indeed, just start at 0 like decent human beings.

PR #23, opened 2018-07-01
Update issue templates

legopiraat:
Added a generic issue template that could be used to file bug reports.


hugo-vrijswijk:
Can you change "verison" to version near the end of the file, and then it should be good

PR #24, opened 2018-07-02
Start mutant ID's from 0

hugo-vrijswijk:
Starts the mutant ID's from 0. Doesn't really make a functional difference, and makes more sense in the code.


hugo-vrijswijk:
Fixed

legopiraat:
Fixes: #24

PR #25, opened 2018-07-02
Rename Sentinel files to Stryker4s

hugo-vrijswijk:
It seems that in renaming from Sentinel to Stryker4s I forgot to rename some files. This PR fixes that.


legopiraat:
On a sidenote please refere to the Issue it fixes. This will auto close it when merged :)

e.g.
Fixes: #25

hugo-vrijswijk:
There's no issue for this as I didn't really see it necessary to make a separate issue for it

PR #32, opened 2018-07-06
Add Better code badge

legopiraat:
I've added `stryker4s` to the `bettercodehub` so we can get some general static code analysis. It should be triggered when a new branch is build/merged i think. If not we can look for a replacement that does so. 

Alternative could be `Codecov` for example


hugo-vrijswijk:
Looks good! An alternative could be [Codacy ](https://www.codacy.com/)(pretty good support for Scala from what I can see), and for code coverage we could use Coveralls. Though I'm not sure we'd want code coverage as a requirement for new PR's

legopiraat:
No i wouldn't want it as a requirement but we should have insights on it :) 

`Codacy` is also a nice product for sure, we could always investigate in which of the two gives more insights.

Issue #29, opened 2018-07-06
Comply with the new HTML reporter specification


legopiraat:
Things that need to be done to comply with new HTML reporter specification so we can integrate with it smoothly.

- [ ] Change internal structure of `MutantRunResults` and `MutantRunResult` to the new report data. 
- [ ] Calculate the span of the mutation.
- [ ] Collect information about mutations per file.
- [ ] Calculate `health`

Followup on this will be to integrate the new html reporter.

legopiraat:
This change seems to have a fairly big impact on the code. Also with several ways to go maybe we should have a discussion on which way we want to go for this change @hugo-vrijswijk.

hugo-vrijswijk:
If it is too much changes in the current codebase, we could always make a mapper or something to leave the current codebase intact. But I think it should be possible

Regarding the mutations per file, that shouldn't be too much work. Right now all mutations are already collected per file, but are then flatmapped when running them. 

The health seems related to # . We could hard-code the health percentages. Though I don't think just implementing health information and a  config without 'breaking' the build on a lower score would be too big of a feature to implement first.

Why did you already check off the span point? Maybe we should do individual PR's for each point. To keep them small.

legopiraat:
At this point i think a mapper is the way to go for now. And in the future work towards a more maintable solution.

nicojs:
Is there any progress on this? There is a [`0.0.4` version of the mutation testing elements](https://github.com/stryker-mutator/mutation-testing-elements/issues/2#issuecomment-464881050). It would be awesome to have an example of a Scala report to test the with.

hugo-vrijswijk:
Yes! I've started working on it again in the `[Comply-html-reporter-format](https://github.com/stryker-mutator/stryker4s/tree/Comply-html-reporter-format)` branch. Believe me, we are just as excited to have something to test with 😀

PR #35, opened 2018-08-09
WIP: Comply html reporter format

legopiraat:
Created a mapper that will map the current run results to the json format from the mutation testing elements. 

Status:
Still very much work in progress and am still missing some of the fields. But the first setup has been made.

### DIFF
--- a/core/src/main/scala/stryker4s/run/report/HtmlRunReporter.scala
+++ b/core/src/main/scala/stryker4s/run/report/HtmlRunReporter.scala
@@ -1,0 +1,13 @@
+package stryker4s.run.report
+import stryker4s.config.Config
+import stryker4s.model.MutantRunResults
+import stryker4s.run.report.mapper.MutantRunResultMapper
+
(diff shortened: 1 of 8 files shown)


hugo-vrijswijk:
Looks like a good start. Did you give changing the current model to more closely match the HTML reporter model a try? Though i imagine not everything in it is applicable.

@nicojs Right now we have a console-reporter. But Stryker4s only calls a single one. That would be something we'd have to change too.

legopiraat:
Good suggestion about the HTML-reporter feature branch, will do that no problem. 

I've looked into changing the current model just for a bit and ended up in a limbo where i changed everything. I'll try to change a few small things here and their but for now i think the mapper will do the majority.

legopiraat:
Create a branch called `HTML-Reporter` which will be the main branch for this new feature.

hugo-vrijswijk:
Closed. We'll move the discussion into the issue #29 or Gitter. A PR can be made when code is ready to be merged

PR #38, opened 2018-08-13
Make process runner configurable

### LINKED ISSUE #36: Configurable ProcessRunner
We should make the ProcessRunner configurable for multiple test commands, instead of a hardcoded `sbt test`. This means we, in principle would, support gradle and maven. 

When in the future we have 'proper' support for a build tool (like SBT in #2 ) we could show a warning when a user runs with the 'slower' process runner.

### LINKED ISSUE #37: Configurable reporter
Ability to configure which reporter(s) should be used after the mutation test run is completed.

I think one or more reporters will be a nice addition.
Shouldn't be hard to implement but on the console you will see a quick mutation score with a duration. And the actual html reporter would contain detailed information.

Merge into: `HTML-Reporter`

hugo-vrijswijk:
Fixes #36

Added the option to configure the test command that the ProcessRunner runs. It can be configured with a `command-runner` test-runner. Future test-runners could be configured in a way similar to Stryker's test-runners and read in the TestReader. 

I added a new TestRunner config value instead of adding the correct MutantRunner as it made more sense to me to keep configuration in the Config object instead of adding an implementation to it. I am not sure about the naming of the classes yet, I think they are confusing. But also don't know what would be better.


legopiraat:
Somehow I feel icky about the `command` because if the user defines the full command we will not be able to preserve the session for example `sbt` which we would like to do. I think I would be more of a fan to split of the process runner command from the other command, something like:

```hocon
{
    command-runner: {
          runner: "sbt"
          command: "test -DskipStuff=true"
     }
 }
```

For sure this will make it usable for all build tools but it has some big downsides.

nicojs:
Great that you're using the same name here: `command-runner` is also the name Stryker uses

hugo-vrijswijk:
@legopiraat I imagine with a SBT (or other) plugin the syntax would be like this, as it won't be a command-runner:
```conf
test-runner: "sbt"
```

I'm not opposed to splitting the command and args into separate values. That would also make it easier to put out a warning when we have a plugin to use instead.

legopiraat:
If you don't mind I think it would be a good idea to split it up. And indeed it will be easier for us to putt out the notifications I suppose a lot of users would appreciate that.

hugo-vrijswijk:
I've split the command and args into separate values

hugo-vrijswijk:
Turns out Pureconfig also has a ConfigWriter function, for writing objects to Typesafe config. I've implemented this, which also fixes #39. I've also removed the TestRunnerReader completely, though the configuration is slightly different. Configuring the TestRunner would now look like this: 
```conf
test-runner: {
  type=commandrunner
  command: "mvn"
  args: "clean test" 
}
```
I'm not completely opposed to the `type` argument, as I think it's still clear what the option does.

PR #46, opened 2018-08-31
Add configuration setting so a reporter can be configured

### LINKED ISSUE #37: Configurable reporter

legopiraat:
Added a way to configure a reporter. 

There is a conversion in place so we know if an incorrect reporter is configured while reading in the configuration. 

Fixes #37


core/src/test/scala/stryker4s/config/ConfigReaderTest.scala
```diff
@@ -26,6 +31,15 @@ class ConfigReaderTest extends Stryker4sSuite with BeforeAndAfterEach {
       exc.getMessage() should include("Key not found: 'stryker4s'.")
     }
 
+    it("should fail on an unknown reporter") {
+      val confPath = FileUtil.getResource("stryker4sconfs/wrongReporter.conf")
+
+      lazy val result = ConfigReader.readConfig(confPath)
+      val exc = the[ConfigReaderException[_]] thrownBy result
+
+      exc.getMessage() should include("Cannot convert configuration")
```
hugo-vrijswijk: Can you also test if the ConfigReader fails on the correct line? (The one with the nonexistent reporter). Should be in the ConfigReaderFailure object off the top of my head

core/src/main/scala/stryker4s/config/Config.scala
```diff
@@ -5,13 +5,18 @@ import java.nio.file.Path
...
+import stryker4s.run.report.{ConsoleReporter, MutantRunReporter}
 
 case class Config(files: Seq[String] = Seq("**/main/scala/**/*.scala"),
                   baseDir: File = File.currentWorkingDirectory,
-                  testRunner: TestRunner = CommandRunner("sbt", "test")) {
+                  testRunner: TestRunner = CommandRunner("sbt", "test"),
+                  reporters: List[MutantRunReporter] = List(new ConsoleReporter)) {
 
   def toHoconString: String = {
-    implicit val writer: ConfigWriter[File] = ConfigWriter[Path].contramap[File](c => c.path)
```
hugo-vrijswijk: As discussed, move this to a companion object

Issue #41, opened 2018-08-24
Improve console output


hugo-vrijswijk:
The console output could be clearer. I think we should try to improve it:

 - ~~ℹ Configurable log levels (right now debug is always on)~~ Fixed #49  
- ~~👓 Improve readability with a nicer logging template with a `logback.xml`. Maybe something similar to Stryker. Personally, I think the long package names are quite ugly and don't add much (who has duplicate class names in different packages anyway?).~~ Fixed with #54 
- 🌈 Colorize some output! (like mutations, log levels, files, etc). [Fansi](https://github.com/lihaoyi/fansi) seems like a good candidate. Ideally, we'd also only output color codes if stdout supports it (so not when outputting to a file, just text terminal etc)
- 📊 Progress bar while running. Also similar to what Stryker has
- ⌚ ETA until mutation run is finished

More ideas are welcome!

legopiraat:
When calculating the ETA we need to keep in mind that the first run will always be slower by a lot.

The reason for is the first time compilation the `Scala` compiler has to do. If we want a more accurate ETA we should base it on the second test run.

cbrunnkvist:
In my experience, the most frustrating thing is not the lack of a somewhat accurate ETA (#267), but rather the absence of _any_ progress report. At the current state and default level, I sometimes end up staring at:`
```
[info] Creating 4 test-runners
[info] Starting initial test run...
```

...for about an hour, with no clue whatsoever as to whether something is hanging or there is some error I should know of or where if there is one specific test / mutant that is responsible the majority of the test-time. Just some "$n/$total" would already be quite an improvement, the name of the test classes running, by each runner, would be just bonus...

cbrunnkvist:
(On a related note: the next logLevel up from Info is obviously Debug, but that one is currently almost _too_ "debuggy" - I think I remember that it is due to SBT's `Level`-enum deciding for all of us that "Trace" is a superfluous level, so I don't think there is any way to improve on logLevel side...)

hugo-vrijswijk:
@cbrunnkvist is there any logging in debug that you think is too debuggy? We can maybe remove some statements if they don't add much

cbrunnkvist:
@hugo-vrijswijk I'm sure we can figure out some pragmatic compromise: like, if we have a flag to leave temporary files / mutation directories around, that flag could also be the one which enables printing "copying XXX into YYY" for example. 

[...]

PR #49, opened 2018-08-31
Allow overriding the log level from the config file

philippw:
You now can set the log level from the stryker4s.conf file using the `log-level` property

Accepted values are ERROR, WARN, INFO, DEBUG, TRACE, ALL, OFF

Fixes bullet point 1 in #41


util/src/test/scala/stryker4s/scalatest/LogMatchers.scala
```diff
@@ -33,12 +33,14 @@ trait LogMatchers {
   def loggedAsInfo = new LogMatcherWithLevel(Level.INFO)
   def loggedAsWarning = new LogMatcherWithLevel(Level.WARN)
   def loggedAsError = new LogMatcherWithLevel(Level.ERROR)
+  def notLoggedAtAll = new LogMatcherWithLevel(Level.OFF)
```
hugo-vrijswijk: How can a message be logged with log level OFF?

core/src/main/scala/stryker4s/run/process/ProcessRunner.scala
```diff
@@ -11,8 +11,8 @@ import scala.util.Try
 
 trait ProcessRunner extends Logging {
   def apply(command: Command, workingDir: File, envVar: (String, String)): Try[Int] = {
-    val mutantProcess = Process(command.command + " " + command.args, workingDir.toJava, envVar)
-      .run(ProcessLogger(s => debug(s)))
+    val mutantProcess = Process(Seq(command.command, command.args), workingDir.toJava, envVar)
```
hugo-vrijswijk: Why is this changed? Wrong merge?

core/src/main/scala/stryker4s/config/Config.scala
```diff
@@ -3,15 +3,22 @@ package stryker4s.config
...
 import better.files._
+import ch.qos.logback.classic.Level
 import com.typesafe.config.ConfigRenderOptions
 import pureconfig.ConfigWriter
 
 case class Config(files: Seq[String] = Seq("**/main/scala/**/*.scala"),
                   baseDir: File = File.currentWorkingDirectory,
-                  testRunner: TestRunner = CommandRunner("sbt", "test")) {
+                  testRunner: TestRunner = CommandRunner("sbt", "test"),
+                  logLevel: Level = Level.DEBUG) {
```
hugo-vrijswijk: Can we change the default log level to INFO? Makes more sense to me, and it's also what Stryker does.

philippw: Yes, I'll do that.

PR #54, opened 2018-09-01
Add a more compact logback.xml than the default one

philippw:
Produces log output like the following:

```
15:05:28.776 grizzled.slf4j.Logger: Setting logging level to INFO. 
15:05:32.142 grizzled.slf4j.Logger: Found 16 of 36 file(s) to be mutated. 
15:05:32.145 grizzled.slf4j.Logger: 81 Mutant(s) generated 
15:05:32.678 grizzled.slf4j.Logger: Starting test-run 0... 
```

### DIFF
--- a/core/src/main/scala/stryker4s/config/Config.scala
+++ b/core/src/main/scala/stryker4s/config/Config.scala
@@ -12,7 +12,7 @@ case class Config(files: Seq[String] = Seq("**/main/scala/**/*.scala"),
                   baseDir: File = File.currentWorkingDirectory,
                   testRunner: TestRunner = CommandRunner("sbt", "test"),
                   reporters: List[MutantRunReporter] = List(new ConsoleReporter),
-                  logLevel: Level = Level.INFO) {
+                  logLevel: Level = Level.DEBUG) {
 
   def toHoconString: String = {
     implicit val fileWriter: ConfigWriter[File] = ConfigWriter[Path].contramap[File](_.path)
--- a/core/src/main/scala/stryker4s/config/Config.scala
+++ b/core/src/main/scala/stryker4s/config/Config.scala
@@ -12,7 +12,7 @@ case class Config(files: Seq[String] = Seq("**/main/scala/**/*.scala"),
                   baseDir: File = File.currentWorkingDirectory,
                   testRunner: TestRunner = CommandRunner("sbt", "test"),
                   reporters: List[MutantRunReporter] = List(new ConsoleReporter),
-                  logLevel: Level = Level.DEBUG) {
+                  logLevel: Level = Level.INFO) {
 
   def toHoconString: String = {
     implicit val fileWriter: ConfigWriter[File] = ConfigWriter[Path].contramap[File](_.path)
--- a/core/src/main/scala/stryker4s/config/ConfigReader.scala
+++ b/core/src/main/scala/stryker4s/config/ConfigReader.scala
@@ -4,8 +4,9 @@ import java.io.FileNotFoundException
 import java.nio.file.Path
 
 import better.files.File
-import ch.qos.logback.classic.Level
+import ch.qos.logback.classic.{Level, Logger}
 import grizzled.slf4j.Logging
+import org.slf4j.{LoggerFactory, Logger => Slf4jLogger}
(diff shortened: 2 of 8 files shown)


hugo-vrijswijk:
Nice! Definitely a lot clearer. Though I think we should also log the log-level

hugo-vrijswijk:
I also noticed that no output is logged before the log level is set with this logback.xml. Not sure what setting in the file causes it

legopiraat:
because by default the rootappender is set to error i think

```
  <root level="error">
        <appender-ref ref="STDOUT"/>
    </root>
```
try changing that to info

legopiraat:
We might have to change the `setLoggingLevel(config.logLevel)` location from `Stryker4sRunner` to the config reader. Because currently either nothing is logged if we set the rootlogger to error/warning. But if we putt it default on debug it's awkward that you see a debug log statement when you for example, configure a `stryker4s.conf` to have log level info and up. 

We could do it as first thing in the `readConfig` or `tryRecoverFromFailures` that way we can also ignore the debug log if the user actually sets it higher then debug.

legopiraat:
@hugo-vrijswijk I changed the default log level to debug otherwise the print of the default config is useless. I think in general the users will setup a `Stryker4s.conf` anyway and can change the log level accordingly if they need to.

core/src/main/scala/stryker4s/config/Config.scala
```diff
@@ -12,7 +12,7 @@ case class Config(files: Seq[String] = Seq("**/main/scala/**/*.scala"),
                   baseDir: File = File.currentWorkingDirectory,
                   testRunner: TestRunner = CommandRunner("sbt", "test"),
                   reporters: List[MutantRunReporter] = List(new ConsoleReporter),
-                  logLevel: Level = Level.INFO) {
+                  logLevel: Level = Level.DEBUG) {
```
hugo-vrijswijk: With the logging level being set right after it's read. Should we still have the default debug? I think a default of info makes more sense

core/src/main/scala/stryker4s/config/ConfigReader.scala
```diff
@@ -47,4 +52,15 @@ object ConfigReader extends Logging {
...
       throw ConfigReaderException(failures)
   }
+
+  /**
+    * Sets the logging level to one of the following levels:
+    * OFF, ERROR, WARN, INFO, DEBUG, TRACE, ALL
+    *
+    * @param level the logging level to use
+    */
+  private def setLoggingLevel(level: Level): Unit = {
```
hugo-vrijswijk: It's a bit ugly to have the log level being a side-effect in the ConfigReader, imo. But it's fine for now

core/src/main/scala/stryker4s/config/Config.scala
legopiraat: The reason why I made the default DEBUG is because of config reader. In that function we are printing the default config that has been used: `debug("Config used: " + defaultConf.toHoconString)` if we set the default to INFO this log statement will never be able to print. 

We could make the default INFO but then I would like to have the config print info as well.

What would you suggest?

core/src/main/scala/stryker4s/config/ConfigReader.scala
legopiraat: i know but otherwise we lose all the config logging

core/src/main/scala/stryker4s/config/Config.scala
legopiraat: well the thing you're missing this, it's all happing inside `tryRecoverFromFailures` which only goes off when there is no `stryker4s.conf` available. So it will purely rely on the defaults set in the config class. Which would be loglevel INFO and thus never print.

hugo-vrijswijk: Ah. Yeah I missed that haha. In that case we could change the log to info. Or maybe we should move the used config to always be logged (at debug) and just accept that it won't be logged in that case.

legopiraat: For now i changed it to INFO and also logged the default config to INFO. I liked that idea the best.

PR #58, opened 2018-09-09
Logging tests without race conditions

### LINKED ISSUE #44: Logging tests race condition?
The logging tests are failing randomly sometimes on the build-server. I think there's a race condition/timing issue in the current code. SBT runs tests in parallel, and I think resetting the test log causes the issue.
Here's what I think happens:

TEST A: Starts test
TEST B: Starts test
TEST A: Logs a statement
TEST B: Logs a statement
TEST A: Assert for logged statement
TEST A: Finished, **resets TestAppender**. Test succeeds!
TEST B: Assert for logged statement. Not found in the TestAppender. Test fails!

legopiraat:
I rewrote the logMatcher + appender so it's safe to use.

Basically how it works is by adding a class name so the reset function only removes log events for that specific class under test. because of the AfterEach which is used to reset this is perfectly safe. 

Also altered the log matchers to include the class name so if 2 different classes would log the same line the matchers would still match correctly. This is also done by including the class name.

One warning: It does require a 1:1 mapping of the package structure for the tests. But this is a good practice to do anyway.

this will fix #44

### DIFF
--- a/core/src/test/scala/stryker4s/config/ConfigReaderTest.scala
+++ b/core/src/test/scala/stryker4s/config/ConfigReaderTest.scala
@@ -93,6 +93,6 @@ class ConfigReaderTest extends Stryker4sSuite with BeforeAndAfterEach {
   }
 
   override def afterEach(): Unit = {
-    TestAppender.reset()
+    TestAppender.reset
   }
 }
--- a/core/src/test/scala/stryker4s/mutants/MutatorTest.scala
+++ b/core/src/test/scala/stryker4s/mutants/MutatorTest.scala
@@ -1,13 +1,13 @@
-package stryker4s
+package stryker4s.mutants
 
 import org.scalatest.BeforeAndAfterEach
 import stryker4s.config.Config
-import stryker4s.mutants.Mutator
 import stryker4s.mutants.applymutants.{MatchBuilder, StatementTransformer}
 import stryker4s.mutants.findmutants.{MutantFinder, MutantMatcher}
 import stryker4s.run.MutantRegistry
 import stryker4s.scalatest.{FileUtil, TreeEquality}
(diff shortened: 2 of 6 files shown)


hugo-vrijswijk:
Cool! Nice that it doesn't need any changed test code. As far as I can tell it doesn't fix it fully, though. There's now fewer resets, but the case described in #44 can still happen, as they are logged by the same class. With fewer resets, there should be a lower chance of it happening, though

legopiraat:
I think this case is actually covered because only the log events for a specific class are removed in the reset.

First this will make a list of all the log events for the specific class. And secondly remove all those log events from all the log events.
`events --= events.filter(event => event.getLoggerName.contains(loggerClassName))`

This should be enough to make sure the scenario from #44 is safe it not? Or am I missing something?

hugo-vrijswijk:
If the logging tests from #44 are testing the same class (which is quite common), then the statements would still be reset, right?

legopiraat:
Given:

TEST A: Starts test
TEST B: Starts test

State: Seq()

TEST A: Logs a statement
TEST B: Logs a statement

State: Seq(Logevent(A), Logevent(B))

TEST A: Assert for logged statement
TEST A: Finished, resets TestAppender. Test succeeds!
TEST A: AfterEach() { TestAppender.reset } <-- Note

State: Seq(Logevent(B))

TEST B: Assert for logged statement. 
TEST B: Finished, resets TestAppender. Test succeeds!
TEST B: AfterEach() { TestAppender.reset } <-- Note

State: Seq()

this is the behavior that is implemented.

hugo-vrijswijk:
Hmm, probably. I've tried running the tests multiple times again and can't seem to reproduce it anymore. So I'll merge this!

PR #63, opened 2018-09-19
Refactor all logging test code into LogMatcher

hugo-vrijswijk:
Refactors the logging test code into `LogMatcher`. This way, tests only have to extend `LogMatcher` to test logs, instead of also having to extend `BeforeAndAfterEach`. Tests now also don't have to call the `afterEach` to reset logs, but this is done automatically when extending `LogMatcher`. Also, because the implicit className is no longer in `Stryker4sSuite`, compile times are theoretically a little faster.

### DIFF
--- a/util/src/test/scala/stryker4s/scalatest/LogMatchers.scala
+++ b/util/src/test/scala/stryker4s/scalatest/LogMatchers.scala
@@ -2,11 +2,10 @@ package stryker4s.scalatest
 
 import ch.qos.logback.classic.Level
 import ch.qos.logback.classic.spi.ILoggingEvent
-import org.scalatest.BeforeAndAfterEach
 import org.scalatest.matchers.{BeMatcher, MatchResult}
 import stryker4s.{Stryker4sSuite, TestAppender}
 
-trait LogMatchers extends BeforeAndAfterEach {
+trait LogMatchers {
   // Causes a compile error if LogMatchers is used without Stryker4sSuite
   this: Stryker4sSuite =>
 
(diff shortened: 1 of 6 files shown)


util/src/test/scala/stryker4s/scalatest/LogMatchers.scala
```diff
@@ -2,10 +2,26 @@ package stryker4s.scalatest
 
 import ch.qos.logback.classic.Level
 import ch.qos.logback.classic.spi.ILoggingEvent
+import org.scalatest.BeforeAndAfterEach
 import org.scalatest.matchers.{BeMatcher, MatchResult}
-import stryker4s.TestAppender
+import stryker4s.{Stryker4sSuite, TestAppender}
 
-trait LogMatchers {
+trait LogMatchers extends BeforeAndAfterEach {
```
legopiraat: Not a fan of having to extend `BeforeAndAfterEach` in the `LogMatchers`  if you need it then maybe place it inside the `Stryker4sSuite` because now you will sometimes have before and after each available if you do log testing. In my opinion that makes no sense, it makes more sense to always have it inside the `Stryker4sSuite` in this case.

util/src/test/scala/stryker4s/scalatest/LogMatchers.scala
```diff
@@ -2,10 +2,26 @@ package stryker4s.scalatest
...
+  def loggedAsWarning = new LogMatcherWithLevel(Level.WARN)
+  def loggedAsError = new LogMatcherWithLevel(Level.ERROR)
+
+  override def afterEach(): Unit = TestAppender.reset
+
+  /**
+    * The className of the system under test.
+    * Is done by getting the executing test class and stripping off 'test'.
+    */
+  implicit val className: String = getClass.getCanonicalName.replace("Test", "")
```
legopiraat: Should be private

hugo-vrijswijk:
@legopiraat Removed the cache and restarted the build. It's now green

PR #66, opened 2018-09-21
Improve calculation of mutation score

legopiraat:
Currently, when the configuration is wrong and you will not find any mutations the mutation score calculation will do a divide by zero. 

Scala's behavior is not a DivideByZeroException by a NaN. This merge request will putt a check if it's a `NaN` and convert that back to a `0.00`

### DIFF
--- a/util/src/main/scala/stryker4s/extensions/score/MutationScoreCalculator.scala
+++ b/util/src/main/scala/stryker4s/extensions/score/MutationScoreCalculator.scala
@@ -1,0 +1,13 @@
+package stryker4s.extensions.score
+
+trait MutationScoreCalculator {
+
+  def calculateMutationScore(totalMutants: Double, detectedMutants: Double): Double = {
+    detectedMutants / totalMutants * 100 match {
+      case mutationScore if mutationScore.isNaN =>
+        0.00
+      case mutationScore =>
+        BigDecimal(mutationScore).setScale(2, BigDecimal.RoundingMode.HALF_UP).toDouble
+    }
+  }
+}
--- a/util/src/test/scala/stryker4s/extensions/score/MutationScoreCalculatorTest.scala
+++ b/util/src/test/scala/stryker4s/extensions/score/MutationScoreCalculatorTest.scala
@@ -31,5 +31,11 @@ class MutationScoreCalculatorTest extends Stryker4sSuite {
 
       mutationScore shouldBe 85.71
     }
+
+    it("Should give a mutation score of 66.67 when 6 of the 9 mutations are killed") {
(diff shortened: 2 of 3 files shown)


util/src/main/scala/stryker4s/extensions/score/MutationScoreCalculator.scala
```diff
@@ -0,0 +1,13 @@
+package stryker4s.extensions.score
+
+trait MutationScoreCalculator {
+
+  def calculateMutationScore(totalMutants: Double, detectedMutants: Double): Double = {
+    detectedMutants / totalMutants * 100 match {
+      case mutationScore if mutationScore.isNaN =>
+        0.00
```
hugo-vrijswijk: We might also want to log something about having zero mutants. It's a valid use case (dry-run, for example), but not something that should happen without the user being unaware

legopiraat: Was thinking about this as well, but i don't think this is the right place for that and maybe not even the correct merge request.

util/src/test/scala/stryker4s/extensions/score/MutationScoreCalculatorTest.scala
```diff
@@ -0,0 +1,35 @@
...
+      val mutationScore = sut.calculateMutationScore(100, 100)
+
+      mutationScore shouldBe 100.00
+    }
+
+    it("Should give a mutation score of 85.71 when 12 of the 14 mutations are killed") {
+      val mutationScore = sut.calculateMutationScore(14, 12)
+
+      mutationScore shouldBe 85.71
+    }
```
hugo-vrijswijk: Can you also add a test for rounding up?

legopiraat: What do you mean round up? We always check on 2 decimals.

hugo-vrijswijk: In this example the result is 85.714. I think we should also test if rounding up works correctly (6/9 to 66.67, for example)

legopiraat: I've added the test but I don't see much value in it since we are testing internal bigdecimal rounding code with this. Which is not our job :)

hugo-vrijswijk: We're not testing BigDecimal, imo. We're testing if our class rounds up. Whether that uses BigDecimal rounding or something else doesn't matter

PR #69, opened 2018-09-25
Fix literal pattern match not being included in top statement

### LINKED ISSUE #48: Unable to mutate primitive types in pattern match
### Summary

Currently Stryker4s is unable to mutate primitive types in a pattern match because it will try to replace it with a new pattern match.

example:
```scala
  def test(): Unit = {
    val a = "hello"

    a match {
      case "world" => println("hello")
      case "hello" => println("world")
    }
  }
```
Will break because it will try to mutate it to
```scala
  def test(): Unit = {
    val a = "hello"

    a match {
      case sys.env.get("ACTIVE_MUTATION") match {
                     case Some("0") => ""
                     case _ => "hello"
                 } => println("hello")
      case "hello" => println("world")
    }
  }
```

hugo-vrijswijk:
Fixes #48 

Also cleaned up the syntax of the `ImplicitTermExtensions` a little bit. Hopefully it's a little clearer. `||` and `&&` (and other infix apply's) are now also included in the top statement, which might make the resulting pattern match a little bigger, but will likely prevent some other compile errors

### DIFF
--- a/core/src/test/scala/stryker4s/mutants/applymutants/MatchBuilderTest.scala
+++ b/core/src/test/scala/stryker4s/mutants/applymutants/MatchBuilderTest.scala
@@ -40,7 +40,7 @@ class MatchBuilderTest extends Stryker4sSuite with TreeEquality {
       implicit val ids: Iterator[Int] = Iterator.from(0)
       val source = "class Foo { def bar: Boolean = 15 > 14 }".parse[Source].get
 
-      val transformed = toTransformed(source, q">", q"<", q"==")
+      val transformed = toTransformed(source, q">", q"<", q"==", q">=")
       val transStatements =
         SourceTransformations(source, List(transformed))
       val sut = new MatchBuilder
@@ -56,6 +56,8 @@ class MatchBuilderTest extends Stryker4sSuite with TreeEquality {
           |      15 < 14
           |    case Some("1") =>
           |      15 == 14
+          |    case Some("2") =>
+          |      15 >= 14
           |    case _ =>
           |      15 > 14
           |  }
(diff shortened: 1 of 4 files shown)


legopiraat:
Does this also fixes String lists not going back to top statements?

(Give it a try with this: `List("a", "b", "c")`

hugo-vrijswijk:
What do you mean? Trying with that statement and matching on a String gives back the whole List statement

legopiraat:
Currently:

List("a", "b", "c") will give:
```scala
activemutant match {
 case "0" => List("", "b", "c")
case _ => activemutant match {
  case "1" => List("a", "", "c")
   case _ => activemutant match {
       case "2" => List("a", "b", "")
       case _ =>List("a", "b", "c")
}}
```

core/src/test/scala/stryker4s/mutants/applymutants/MatchBuilderTest.scala
```diff
@@ -79,21 +79,21 @@ class MatchBuilderTest extends Stryker4sSuite with TreeEquality {
       // Assert
       val expected =
         """class Foo {
-          |  def bar: Boolean = (sys.env.get("ACTIVE_MUTATION") match {
+          |  def bar: Boolean = sys.env.get("ACTIVE_MUTATION") match {
           |    case Some("0") =>
-          |      15 < 14
+          |      15 < 14 && 14 >= 13
```
legopiraat: Am missing a mutation here:

`case GreaterThan(orig)        => orig ~~> (GreaterThanEqualTo, LesserThan, EqualTo)` 
Where is the `GreaterThanEqualTo` ?

core/src/test/scala/stryker4s/mutants/applymutants/MatchBuilderTest.scala
```diff
@@ -79,21 +79,21 @@ class MatchBuilderTest extends Stryker4sSuite with TreeEquality {
...
-          |      14 == 13
+          |      15 == 14 && 14 >= 13
           |    case _ =>
-          |      14 >= 13
-          |  })
+          |      sys.env.get("ACTIVE_MUTATION") match {
+          |        case Some("2") =>
+          |          15 > 14 && 14 > 13
+          |        case Some("3") =>
+          |          15 > 14 && 14 == 13
```
legopiraat: Missing a mutation here as well

`case GreaterThanEqualTo(orig) => orig ~~> (GreaterThan, LesserThan, EqualTo)`

core/src/test/scala/stryker4s/mutants/applymutants/MatchBuilderTest.scala
hugo-vrijswijk: The MutantMatcher wasn't being used here, so the mutations were added above. I've changed it to three to reflect the amount that'd show up in a real case, though

core/src/test/scala/stryker4s/mutants/applymutants/MatchBuilderTest.scala
hugo-vrijswijk: Same, see above

PR #71, opened 2018-09-29
Cleanup tmp files after mutation run completes or crashes

### LINKED ISSUE #53: Cleanup tmp files after mutation run completes or crashes
### Summary

Currently when a test run is completed or crashes the tmp folder which was created is not cleaned.
This is not that user-friendly and we should make sure the files which are created as cleaned.

legopiraat:
Will eventually fix #53 

Just to keep clear on what's happening here I've made this pull request.

- [x] Copy over files only known to git
- [x] Fallback for when it's not a git repo (should be the files configuration setting)
- [x] Cleanup code

I did not feel the need to implement an actual clean up because everything is in the target folder anyway. This can be cleaned easily using the build tool or simply delete the target folder.


core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
```diff
@@ -1,29 +1,80 @@
...
       .filter(file => file.startsWith("!"))
-      .flatMap(file => config.baseDir.glob(file.stripPrefix("!")))
-      .distinct
+      .map(file => file.stripPrefix("!")))
+  }
+
+  /**
+    * List of all previously copied files if the target folder is not cleaned.
+    */
+  private[this] val stryker4sTmpFiles: Seq[File] = {
```
hugo-vrijswijk: I'm a bit confused about why this is needed. The tmp files should already be filtered, and I don't think it's the current processes responsibility to clean up a previous run. (Also isn't it cleaned by running `sbt clean`?)

legopiraat: Indeed you can clean with `sbt clean` but if you don't do that old stryker runs are included. 
And if the user doesn't specify the `mutate` options correctly so it doesn't match target files. Old stryker run files are included and also mutated again.

core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
```diff
@@ -1,29 +1,80 @@
...
+
+  private[this] def fallBackToFilesConfiguration(): Seq[File] = {
+    config.files match {
+      case Some(files) => glob(files)
+      case None =>
+        warn("No 'files' specified falling back to copying everything excluding target")
+
+        config.baseDir
+          .glob("**/*.*")
+          .filterNot(file => file.path.toString.contains("target"))
```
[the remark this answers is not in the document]
legopiraat: true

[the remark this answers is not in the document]
legopiraat: fixed

legopiraat:
@hugo-vrijswijk could you check this branch out locally and see how many tests run?

locally all my tests are green:
```
[info] Run completed in 4 seconds, 110 milliseconds.
[info] Total number of tests run: 91
[info] Suites: completed 13, aborted 0
[info] Tests: succeeded 91, failed 0, canceled 0, ignored 0, pending 0
```

on travis one fails:
```
[info] Run completed in 3 seconds, 748 milliseconds.
[info] Total number of tests run: 93
[info] Suites: completed 13, aborted 0
[info] Tests: succeeded 92, failed 1, canceled 0, ignored 0, pending 0
[info] *** 1 TEST FAILED ***
```
and even more fun is that travis somehow got 2 more tests........

hugo-vrijswijk:
Haha, that's awkward. I won't be able to check it until tomorrow evening though

core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
```diff
@@ -1,29 +1,80 @@
...
+
+  private[this] def fallBackToFilesConfiguration(): Seq[File] = {
+    config.files match {
+      case Some(files) => glob(files)
+      case None =>
+        warn("No 'files' specified falling back to copying everything excluding target")
+
+        config.baseDir
+          .glob("**/*.*")
+          .filterNot(file => file.path.toString.contains("/target/"))
```
hugo-vrijswijk: Thinking about it, is the string also a `/` on Windows?

core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
```diff
@@ -1,29 +1,80 @@
...
+        info("Not a git repo falling back to 'files' configuration.")
+        fallBackToFilesConfiguration()
+    }
+  }
+
+  private[this] def fallBackToFilesConfiguration(): Seq[File] = {
+    config.files match {
+      case Some(files) => glob(files)
+      case None =>
+        warn("No 'files' specified falling back to copying everything excluding target")
```
hugo-vrijswijk: Should have a comma after 'specified', and changing the text to something like 'copying everything except for the target/ folder' might be clearer

core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
```diff
@@ -1,29 +1,80 @@
...
+    *
+    * Option 1: Copy every file that is listed by git.
+    * Option 2: Copy every file that is listed by the 'files' config setting
+    * Option 3: Copy every file in the 'baseDir' excluding target folders.
+    */
+  override def filesToCopy(processRunner: ProcessRunner): Iterable[File] = {
+    processRunner(Command("git ls-files", "--others --exclude-standard --cached"), config.baseDir) match {
+      case Success(files) => files.map(config.baseDir / _).distinct
+      case Failure(_) =>
+        info("Not a git repo falling back to 'files' configuration.")
```
hugo-vrijswijk: Same here, comma after 'git repo'.

core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
hugo-vrijswijk: Also see my comment above about the tests

core/src/test/scala/stryker4s/mutants/findmutants/FileCollectorTest.scala
```diff
@@ -215,18 +215,17 @@ class FileCollectorTest extends Stryker4sSuite with MockitoSugar with LogMatcher
       results should contain theSameElementsAs expectedFileList
     }
 
-    it("Should copy all files when no files config key is found and target repo is not a git repo") {
+    it("Should copy not copy files our of the target folders when no files config key is found and target repo is not a git repo") {
```
hugo-vrijswijk: Should copy not copy? 🤔

legopiraat: if only i had a deep learning grid that knows how i write my danglish and fixes it for me automaticly

core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
```diff
@@ -1,29 +1,87 @@
...
-  private[this] val filesToMutate: Seq[File] = {
-    config.mutate
-      .filterNot(file => file.startsWith("!"))
+  /**
+    * Collect all files that are needed to be copied over to the Stryker4s-tmp folder.
+    *
+    * Option 1: Copy every file that is listed by git.
+    * Option 2: Copy every file that is listed by the 'files' config setting
+    * Option 3: Copy every file in the 'baseDir' excluding target folders.
+    */
```
hugo-vrijswijk: I think the order here should be different. If the user has specified `files`, use that. If not, use git, if not, option 3. This way the user can still override any problems they come into with the git listing

core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
```diff
@@ -31,31 +31,47 @@ class FileCollector(implicit config: Config) extends SourceCollector with Loggin
...
-      case None =>
-        warn(
-          "No 'files' specified, falling back to copying everything except the target/ folder(s)")
-
+  /**
+    * List all files from the base directory specified in the Stryker4s basedir config key.
+    */
+  private[this] def listAllFiles(): Option[Iterable[File]] = {
+    Option(
+      (config.baseDir.glob("*.*") ++
```
hugo-vrijswijk: Can't you use `config.baseDir.listRecursively` here instead of two two globs?

legopiraat: i guess that works the same yup

hugo-vrijswijk: Cool! I'll merge once the tests pass

core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
```diff
@@ -1,29 +1,102 @@
...
+
+      listAllFiles()
+    }).getOrElse(Seq.empty)
+  }
+
+  /**
+    * List all files based on the 'files' configuration key from stryker4s.conf.
+    */
+  private[this] def listFilesBasedOnConfiguration(): Option[Iterable[File]] = {
+    debug("Collecting files based on 'files' configuration key.")
```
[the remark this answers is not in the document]
legopiraat: There is no way of logging this after knowing the result without doing some annoying `if`statements. 

So we have 2 options, in my opinion, we either delete the log statements or we let them be there. What would you prefer?

legopiraat:
I did not remove the files on purpose, since they are in the target folder already. Users can clean that easily anyway. 

If you want I can Mark the files to be deleted on Exit and that should work. It wouldn't be too hard to delete all the files if you really want to.

core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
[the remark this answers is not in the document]
hugo-vrijswijk: Can't you do something like 
```scala
val result = config.files.map { glob }
if (result.isDefined) //logging here

result
```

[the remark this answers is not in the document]
legopiraat: We can, i just find it very ugly.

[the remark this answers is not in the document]
hugo-vrijswijk: Haha, I agree. Maybe we should leave it

hugo-vrijswijk:
Hmm, I guess it would be nice to have it for debugging purposes, but I think in normal cases the user won't manually delete it very often (and it can get pretty big while testing a bit). Also, I found that running `sbt clean` doesn't remove the folder

core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
```diff
@@ -1,29 +1,102 @@
...
+        .filter(file => file.startsWith("!"))
+        .map(file => file.stripPrefix("!")))
+  }
+
+  /**
+    * List of all previously copied files if the target folder is not cleaned.
+    */
+  private[this] val stryker4sTmpFiles: Seq[File] = {
+    config.baseDir
+      .glob(s"target${pathSeparator}stryker4s-*")
```
hugo-vrijswijk: I think the `better.files` glob doesn't work with windows path separators, but always uses `/`, so this wouldn't work on windows

Issue #80, opened 2018-10-05
Improve StatementTransformer error logging


hugo-vrijswijk:
If the StatementTransformer is unable to transform a original statement to it's pattern match an unhelpful exception is thrown (see [here](https://github.com/stryker-mutator/stryker4s/issues/57#issuecomment-427472345)). In most cases this is a bug. It'd be great to have some better logging for it. 

What I think could improve:

- Show both the original and to-mutate pattern match syntax, instead of the Scalamets objects
- Show the file and line:column the original statement was in
- As this is likely a bug in Stryker4s, mention that and provide a link to create an issue

legopiraat:
How feasible is the `original and to-mutate pattern match syntax,`?

hugo-vrijswijk:
I think you can just call .syntax on the tree (at least on the original one, maybe not on the one that fails to build)

PR #85, opened 2018-10-09
Improve log messages when a match can not be build

legopiraat:
Fixes: #80 

Improved logging statement. 
Unable to test because there are no known statements at the moment that break so the exception was force tested at run time.

### DIFF
--- a/core/src/main/scala/stryker4s/mutants/applymutants/MatchBuilder.scala
+++ b/core/src/main/scala/stryker4s/mutants/applymutants/MatchBuilder.scala
@@ -1,13 +1,14 @@
 package stryker4s.mutants.applymutants
 
+import grizzled.slf4j.Logging
 import stryker4s.extensions.TreeExtensions.ImplicitTreeExtensions
 import stryker4s.model.{Mutant, SourceTransformations, TransformedMutants}
 
 import scala.meta.contrib.implicits.Equality.XtensionTreeEquality
-import scala.meta.quasiquotes._
-import scala.meta.{Case, Lit, Pat, Term, Tree}
+import scala.meta.{Case, Lit, Pat, Term, Tree, _}
+import scala.util.{Failure, Success}
 
-class MatchBuilder {
+class MatchBuilder extends Logging {
 
   def buildNewSource(transformedStatements: SourceTransformations): Tree = {
     val source = transformedStatements.source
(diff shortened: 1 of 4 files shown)


core/src/main/scala/stryker4s/mutants/applymutants/MatchBuilder.scala
```diff
@@ -20,6 +21,16 @@ class MatchBuilder {
...
+        } match {
+          case Success(value) => value
+          case Failure(exception) =>
+            error(
+              s"Failed to construct pattern match: original statement [$origStatement] " +
+                s"on line [${origStatement.pos.endLine}] for file [${origStatement.pos.input}].)")
+            error(s"Failed mutation(s) ${mutant.mutantStatements.mkString(",")}.")
+            debug(exception.getMessage)
+
+            throw exception
```
hugo-vrijswijk: The original exception gives a *lot* of confusing logging. Maybe we should throw a custom exception here, indicating what happened

core/src/main/scala/stryker4s/mutants/applymutants/MatchBuilder.scala
```diff
@@ -20,6 +21,16 @@ class MatchBuilder {
         rest transformOnce {
           case found if found.isEqual(origStatement) && found.pos == origStatement.pos =>
             buildMatch(mutant)
+        } match {
+          case Success(value) => value
+          case Failure(exception) =>
+            error(
+              s"Failed to construct pattern match: original statement [$origStatement] " +
+                s"on line [${origStatement.pos.endLine}] for file [${origStatement.pos.input}].)")
+            error(s"Failed mutation(s) ${mutant.mutantStatements.mkString(",")}.")
```
hugo-vrijswijk: Some interpunction changes I think look nicer: 

```scala
error(s"Failed to construct pattern match. Original statement: [$origStatement]")
error(s"Failed mutation(s) ......")
error(s"at ${file}....")
```

Also, I think we should use the same notation we use when reporting survived mutants [here](https://github.com/stryker-mutator/stryker4s/blob/master/core/src/main/scala/stryker4s/run/report/ConsoleReporter.scala#L30). And with a relativePath. So something like `src/main/scala/stryker4s/Stryker4sRunner:80:14`

I think we should also log that this is likely a user with Stryker4s, and for the user to enable debug logging and create an issue (with link, when in debug?) with the debugged statement

legopiraat: We can't get the relative path from the data we currently have sadly. That would require us to pass on the original file as well but I fell that's not the way to go just for logging.

core/src/main/scala/stryker4s/mutants/applymutants/MatchBuilder.scala
```diff
@@ -21,16 +22,19 @@ class MatchBuilder extends Logging {
         rest transformOnce {
           case found if found.isEqual(origStatement) && found.pos == origStatement.pos =>
             buildMatch(mutant)
+          throw new Exception("")
```
hugo-vrijswijk: Left from debugging?

legopiraat: yup

PR #91, opened 2018-10-16
Sbt plugin

scelestino:
Now Stryker works directly as a plugin and through `scripted` command

### DIFF
--- a/core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
+++ b/core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
@@ -1,5 +1,6 @@
 package stryker4s.mutants.findmutants
 
+import better.files.File.VisitOptions
 import better.files._
 import grizzled.slf4j.Logging
 import stryker4s.config.Config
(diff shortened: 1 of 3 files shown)


core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
```diff
@@ -56,24 +56,30 @@ class FileCollector(implicit config: Config) extends SourceCollector with Loggin
...
     * List all files from the base directory specified in the Stryker4s basedir config key.
     */
-  private[this] def listAllFiles(): Option[Iterable[File]] = {
-    Option(config.baseDir
+  private[this] def listAllFiles(): Iterable[File] = {
+    config.baseDir
       .listRecursively
-      .filterNot(file => file.pathAsString.contains(s"${pathSeparator}target$pathSeparator"))
-      .toIterable)
+      .filterNot(isTarget)
```
hugo-vrijswijk: Could you move the `filterNot` here and in the other places to `collectFilesToMutate`? I think you could also remove the `stryker4sTmpFiles` as that filters out files that are in the target folder anyway

PR #92, opened 2018-10-16
Sbt plugin

scelestino:
Fix  FileCollector, filesToCopy method (target must to be filtered)


hugo-vrijswijk:
I already just fixed it. But thanks for noticing it too! 😊

PR #94, opened 2018-10-16
Scripted test 2.11

scelestino:
add scripted scala 2.11

remove scala cross version 2.11 (it is not necessary)

fix FileCollector

### DIFF
--- a/core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
+++ b/core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
@@ -41,7 +41,7 @@ class FileCollector(implicit config: Config) extends SourceCollector with Loggin
       warn("No 'files' specified and not a git repository.")
       warn("Falling back to copying everything except the target/ folder(s)")
       listAllFiles()
-    }
+    }.filterNot(isTarget)
   }
 
   /**
--- a/core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
+++ b/core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
@@ -91,8 +91,10 @@ class FileCollector(implicit config: Config) extends SourceCollector with Loggin
     * Is the file in the target folder, and thus should not be copied over
     */
   private[this] def isTarget(file: File): Boolean = {
-    val pathString = file.toString
+    val pathString = file.relativePath.toString
 
+    (file.isDirectory && pathString == "target") ||
+    pathString.startsWith(s"target$pathSeparator") ||
     pathString.contains(s"${pathSeparator}target$pathSeparator") ||
     pathString.endsWith(s"${pathSeparator}target")
   }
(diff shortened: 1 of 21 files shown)


core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
```diff
@@ -91,9 +91,9 @@ class FileCollector(implicit config: Config) extends SourceCollector with Loggin
     * Is the file in the target folder, and thus should not be copied over
     */
   private[this] def isTarget(file: File): Boolean = {
-    val relativePathString = file.relativePath.toString
+    val pathString = file.toString
```
hugo-vrijswijk: This is causing some failing tests. The relative path is done because if the project happens to be somewhere in a folder named target, it should still copy the files

scelestino: The problem with that is that using relativePath the target filter doesn't work, so if you run Stryker twice, the second one will try to mutate the mutated files of the first one.

It is easy to replicate. Using relativePath just run `sbt stryker` twice inside `test-1` project. Do not use `scripted`, run SBT inside test-1 directory like if it was a project that use Stryker plugin.

hugo-vrijswijk: Hmm, I think that would be the starting / in the match. I'll see if I can fix it, because I think the relative method is preferrable

sbt/src/sbt-test/sbt-stryker4s/test-2.11/src/main/scala/example/Example.scala
```diff
@@ -0,0 +1,5 @@
+package example
+
+object Example {
+  def canDrink(age: Int): Boolean = age >= 18
```
hugo-vrijswijk: I think we should also try to ensure the project works with 2.11. I couldn't find any breaking changes between 2.11 and 2.12, but I think we can use the [JavaConversions](https://www.scala-lang.org/api/2.12.3/scala/collection/JavaConversions$.html) deprecated in 2.12 with a scalacOption to not compile with [deprecation errors](https://tpolecat.github.io/2017/04/25/scalac-flags.html) (`-deprecation` and `-Xfatal-warnings`, I think)

core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
scelestino: Oh, I found the issue.

> base dir path
> /home/scelestino/dev/stryker4s2/sbt/src/sbt-test/sbt-stryker4s/test-1
> 
> file path
> /home/scelestino/dev/stryker4s2/sbt/src/sbt-test/sbt-stryker4s/test-1/target/stryker4s-3861747083202168747/src/main/scala/example/Example.scala
> 
> relative
> target/stryker4s-3861747083202168747/src/main/scala/example/Example.scala

Relative doesn't start with slash :confused: 

I'm going to fix isTarget method.

project/Dependencies.scala
```diff
@@ -5,7 +5,8 @@ object Dependencies {
   object versions {
     val scala212 = "2.12.7"
     val scala211 = "2.11.12"
-    val crossScala = Seq(scala211, scala212)
+//    val crossScala = Seq(scala211, scala212)
+    val crossScala = Seq(scala212)
```
hugo-vrijswijk: I think you can remove this, right? It's not used anywhere anymore. If you also update the travis cross-versions I will merge this

PR #96, opened 2018-10-19
Disable mutations

### LINKED ISSUE #4: Disable mutators
For example, disable mutation x because it keeps giving compile errors or false positives. 

Should be easy via `stryker4s.conf` configuration file.
Add documentation on how to exclude mutations.

Example:
```hocon
excludedMutations: ["BooleanSubstitution"]
```

- [ ] Configuration setting
- [ ] document in [configuration.md](https://github.com/stryker-mutator/stryker4s/blob/master/docs/CONFIGURATION.md) for inspiration look at the [original stryker config](https://github.com/stryker-mutator/stryker/blob/master/packages/stryker/README.md#mutator)
- [ ] Exclude mutation when disabled in [mutant matcher](https://github.com/stryker-mutator/stryker4s/blob/master/core/src/main/scala/stryker4s/mutants/findmutants/MutantMatcher.scala)

triggerNZ:
In the spirit of hacktoberfest, this fixes #4 . This is my first PR to this repo so any feedback at all is more than welcome.

### DIFF
--- a/core/src/main/scala/stryker4s/model/Mutant.scala
+++ b/core/src/main/scala/stryker4s/model/Mutant.scala
@@ -2,6 +2,7 @@ package stryker4s.model
 
 import stryker4s.extensions.mutationtypes.Mutation
 
+import scala.language.existentials
 import scala.meta.{Term, Tree}
 
 case class Mutant(id: Int, original: Term, mutated: Term, mutationType: Mutation[_ <: Tree])
--- a/core/src/main/scala/stryker4s/model/Mutant.scala
+++ b/core/src/main/scala/stryker4s/model/Mutant.scala
@@ -1,5 +1,7 @@
 package stryker4s.model
 
-import scala.meta.Term
+import stryker4s.extensions.mutationtypes.Mutation
 
-case class Mutant(id: Int, original: Term, mutated: Term)
+import scala.meta.{Term, Tree}
+
+case class Mutant(id: Int, original: Term, mutated: Term, mutationType: Mutation[_ <: Tree])
--- a/core/src/main/scala/stryker4s/model/Mutant.scala
+++ b/core/src/main/scala/stryker4s/model/Mutant.scala
@@ -2,7 +2,6 @@ package stryker4s.model
 
 import stryker4s.extensions.mutationtypes.Mutation
 
-import scala.language.existentials
 import scala.meta.{Term, Tree}
 
 case class Mutant(id: Int, original: Term, mutated: Term, mutationType: Mutation[_ <: Tree])
--- a/stryker4s.conf
+++ b/stryker4s.conf
@@ -1,4 +1,0 @@
-stryker4s {
-  excluded-mutations = [""]
-
-}
--- a/stryker4s.conf
+++ b/stryker4s.conf
@@ -1,0 +1,4 @@
+stryker4s {
+  excluded-mutations = [""]
+
+}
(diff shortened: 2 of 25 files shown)


hugo-vrijswijk:
Cool! Thank you for your PR! Your code looks mostly good to me. Could you add some tests, so we can ensure it'll keep working in the future? 

Moving the Exclusions to the mutation categories might make it easier to add tests

nicojs:
I like this approach. All mutants are still collected, but filtered later. @hugo-vrijswijk should it be added to a log message? Something like this: "Found x mutant(s) (42 excluded)"? I'm not sure what the current place log message looks like.

legopiraat:
Sounds like a good plan to log it. If we want a full trace we could even log what en were the mutations were excluded but i guess that might be out of scope

hugo-vrijswijk:
I like the idea of logging excluded mutants count! Amount of mutants is currently logged in `Mutator`

https://github.com/stryker-mutator/stryker4s/blob/908d0e5e70f3abf058c297c6b73d4c9750517850/core/src/main/scala/stryker4s/mutants/Mutator.scala#L27-L28

triggerNZ:
I pushed another commit, with a test that checks that exclusions actually apply, and surprise surprise, they didn't. I have fixed it but it requires that the mutation is carried along with the mutant. It doesn't feel like the most elegant thing, but I'm submitting it for feedback from you guys since i'm new to stryker4s.

triggerNZ:
Logging also now done. This will allow us to log individual exclusions too should we choose to in the future.

legopiraat:
Looking into the pull request and some fishy stuff is going on.

```scala
16:08:33.109 INFO  Mutator: Found 19 of 43 file(s) to be mutated.
16:08:33.110 INFO  Mutator: 104 Mutant(s) generated. 11 excluded. <-- BooleanSubstitutions excluded

16:09:09.405 INFO  Mutator: Found 21 of 43 file(s) to be mutated.
16:09:09.407 INFO  Mutator: 118 Mutant(s) generated. 0 excluded. <-- nothing excluded
```

triggerNZ:
Apologies for disappearing for a while. How did you produce the logs above? And I am guessing the bug is that the numbers don't add up? 104 + 11 != 118, yeah? I'll have a look at this this week

hugo-vrijswijk:
No worries, I've been more bit absent than I'd like, myself. If you don't have the time to finish this PR, that's absolutely fine, and we can take it over if you'd like.

I'm guessing the logging above is produced by just running Stryker4s on itself. As a note for the logging, I'd expect the total mutants generated not to change when some are excluded (they're still generated). So, having something like `118 Mutant(s) generated. Out of which 11 excluded`. And perhaps even not show `0 excluded` when nothing is excluded.

legopiraat:
It looks like files are not taken into account at all when all the mutations of that file are excluded. This will mess up with the logging. I' ll take a look if I can add those files as well.

core/src/main/scala/stryker4s/model/Mutant.scala
```diff
@@ -2,6 +2,7 @@ package stryker4s.model
 
 import stryker4s.extensions.mutationtypes.Mutation
 
+import scala.language.existentials
```
hugo-vrijswijk: What's this for?

stryker4s.conf
```diff
@@ -0,0 +1,4 @@
+stryker4s {
+  excluded-mutations = [""]
```
hugo-vrijswijk: Is this file still needed? The default is no exclusions, right?

PR #97, opened 2018-10-19
Add Scalameta cache clean to avoid errors in between sessions

scelestino:
Add a clean to Scalameta cache in order to avoid errors while processing files.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ cache:
 
 scala:
    - 2.12.7
-   - 2.11.12
 
 before_cache:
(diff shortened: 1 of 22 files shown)


hugo-vrijswijk:
Could you put this in the main Stryker4sRunner? It's only run once, so doesn't seem like a huge performance hit. And it might be useful for future plugins as well

scelestino:
> Could you put this in the main Stryker4sRunner? It's only run once, so doesn't seem like a huge performance hit. And it might be useful for future plugins as well

You are right, this can be needed for future plugins.

PR #98, opened 2018-10-27
Create Roadmap

legopiraat:
Initial version for strkyer4s vision, still work in progress (need input from @hugo-vrijswijk as well).

### DIFF
--- a/docs/ROADMAP.md
+++ b/docs/ROADMAP.md
@@ -1,4 +1,4 @@
-# Stryker4s vision of the future
+# Stryker4s, roadmap of the future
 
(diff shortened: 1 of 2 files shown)


nicojs:
Cool. Can we link to this file from the website?

Maybe should be called "roadmap"?

legopiraat:
Yes we should link to this from the website 👍 ( I wanted to start writing something down first and then edit the website.)

> Maybe should be called "roadmap"?
Am fine with that

hugo-vrijswijk:
Cool! The only thing I'm concerned about would be keeping it up to date. And I think it shouldn't reflect the issues backlog too much

PR #101, opened 2018-11-22
Remove cross compilation to scala 2.11

legopiraat:


hugo-vrijswijk:
Some further explanation: 2.11 was giving some issues, and as we will be mostly publishing plugins there's no need for 2.11 cross compilation. We can always add it again later

PR #102, opened 2018-11-22
Wrap Term.Name interpolation statements in a block

### LINKED ISSUE #99: Mutating interpolated strings
[There's a bug in Scalameta](https://github.com/scalameta/scalameta/issues/1792) that we're running into when mutating interpolated strings. Basically, Scalameta doesn't put brackets in statements where it should (instead of `s"${foo}bar"` it puts `s"$foobar"`) after which the compiler complains it can't find the `foobar` variable.

We could fix this either by fixing the bug in Scalameta, or wrapping the interpolated variable in a `Term.Block` when mutating the interpolated string

hugo-vrijswijk:
Fixes #99 

All `Term.Name` statements are now wrapped in `Term.Block`. This only applies to a `Term.Name`, to reduce any changes done to other statements.

I've also added a test to 'prove' the scalameta bug. When it is fixed, the test should fail so we know to remove the workaround.

### DIFF
--- a/core/src/test/scala/stryker4s/mutants/findmutants/MutantMatcherTest.scala
+++ b/core/src/test/scala/stryker4s/mutants/findmutants/MutantMatcherTest.scala
@@ -45,8 +45,11 @@ class MutantMatcherTest extends Stryker4sSuite with TreeEquality {
                         original: MethodMutator,
                         expectedMutations: MethodMutator*): Unit = {
     val found: Seq[Mutant] = tree.collect(matchFun).flatten
-    expectedMutations.foreach {
-      expectedMutation => found.map(_.mutated).collectFirst { case expectedMutation(_, _) => }.getOrElse(fail("mutant not found"))
+    expectedMutations foreach { expectedMutation =>
+      found
+        .map(_.mutated)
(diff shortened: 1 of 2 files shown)


core/src/test/scala/stryker4s/mutants/findmutants/MutantMatcherTest.scala
```diff
@@ -376,14 +386,45 @@ class MutantMatcherTest extends Stryker4sSuite with TreeEquality {
...
+    }
+
+    it("checks if the Scalameta workaround is still needed") {
+      // If this test fails, the bug mentioned above is fixed, and the workaround can be removed
+      val interpolated =
+        Term.Interpolate(q"s",
+                         List(Lit.String("interpolate this"), Lit.String("bar")),
+                         List(q"foo"))
+
+      interpolated.syntax should equal("""s"interpolate this$foobar"""")
```
legopiraat:
```suggestion
      // We expect that after the fix the interpolate string will look as followed.
      // interpolated.syntax should equal("""s"interpolate this${foo}bar"""")
      
      interpolated.syntax should equal("""s"interpolate this$foobar"""")
```

PR #105, opened 2018-11-22
ConfigReader checks if 'excluded-mutations' are valid

Wmaarts:


core/src/test/scala/stryker4s/config/ConfigReaderTest.scala
```diff
@@ -55,16 +55,24 @@ class ConfigReaderTest extends Stryker4sSuite with LogMatchers {
...
+      val confPath = FileUtil.getResource("stryker4sconfs/invalidExcludedMutation.conf")
+
+      lazy val result = ConfigReader.readConfig(confPath)
+      val exc = the[ConfigReaderException[_]] thrownBy result
 
       val head = exc.failures.head
       head shouldBe a[ConvertFailure]
       head.description should equal(
-        s"""No valid coproduct choice found for '{"args":"foo","command":"bar","type":"someOtherTestRunner"}'.""")
+        """Invalid exclusion option(s): 'Invalid, StillInvalid'
```
hugo-vrijswijk: If you mix in the `LogMatchers` trait you can also test the message being logged as `"messageHere" shouldBe loggedAsWarning`

core/src/main/scala/stryker4s/config/implicits/ConfigReaderImplicits.scala
```diff
@@ -16,13 +19,26 @@ trait ConfigReaderImplicits {
...
+      .map(warnInvalidExclusions)
+      .map(exclusions => Exclusions(exclusions.toSet))
+
+  private def warnInvalidExclusions(list: List[String]): List[String] = {
+    val (valid, invalid) = list.partition(Mutation.mutations.contains)
+    if (invalid.nonEmpty) {
+      val errorMessage =
+        s"""Invalid exclusion option(s): '${invalid.mkString(", ")}'
+            |Valid exclusions are ${Mutation.mutations.mkString(", ")}""".stripMargin
+      throw new Exception(errorMessage)
```
hugo-vrijswijk: Maybe a more specific exception would be clearer, and easier to catch if we ever want to catch it in the future (unlikely, but still). There already is a package `stryker4s.extensions.exceptions` where you could put it

legopiraat: I think it should be an extension of `FailureReason` that's the pureconfig way of providing these errors.

[reference](https://pureconfig.github.io/docs/error-handling.html#custom-failure-reasons)

util/src/main/scala/stryker4s/extensions/mutationtypes/MutationTypes.scala
```diff
@@ -7,7 +7,20 @@ import scala.meta.{Lit, Term, Tree}
...
   */
-sealed trait Mutation[T <: Tree]
+sealed trait Mutation[T <: Tree] {
+  val mutationName: String
+}
+
+object Mutation{
+  // List of mutations
+  val mutations: List[String] = List[String](
+    "BinaryOperator",
```
legopiraat: We could also change this static strings to this `classOf[BinaryOperator].getSimpleName` This would be a tiny bit safer when changing the mutator names but I agree the gain is little.

@hugo-vrijswijk @Wmaarts what do you think?

Wmaarts: I thought about it a bit and I agree. Thought about wrapping `classOf[T].getSimpleName` in a method or even in a trait, but that does feel like overthinking / overengineering things to me.

legopiraat: i tried putting it into a method and that's impossible because of type erasure. Let's keep it like this

core/src/main/scala/stryker4s/config/implicits/ConfigReaderImplicits.scala
```diff
@@ -27,18 +28,25 @@ trait ConfigReaderImplicits extends Logging {
...
-  private def warnInvalidExclusions(list: List[String]): List[String] = {
-    val (valid, invalid) = list.partition(Mutation.mutations.contains)
-    if (invalid.nonEmpty) {
-      val errorMessage =
-        s"""Invalid exclusion option(s): '${invalid.mkString(", ")}'
-            |Valid exclusions are ${Mutation.mutations.mkString(", ")}""".stripMargin
-      throw new Exception(errorMessage)
-    }
-    valid
+  private def errorOnInvalidExclusions(configCursor: ConfigCursor): Either[ConfigReaderFailures, List[String]] = {
```
Wmaarts: Here's my attempt at making it more in line with PureConfig. I might need help for this one. :/

core/src/main/scala/stryker4s/config/implicits/ConfigReaderImplicits.scala
Wmaarts: @legopiraat I attempted this [here](https://github.com/stryker-mutator/stryker4s/pull/105/commits/4d84ee4c42692df7d929959b731b2c2457a3fa82#r235942059).

core/src/main/scala/stryker4s/config/implicits/ConfigReaderImplicits.scala
legopiraat: pff, this added a lot more complexity then I expected. I suspected `FailureReason` would extend exception in some sort and then it would be really easy to create a custom exception that extended from FailureReason.

Sorry for this but I think the first solution is a bit more understandable. You could still make a custom exception in the exception package. An example can be found here: https://github.com/stryker-mutator/stryker4s/blob/master/util/src/main/scala/stryker4s/extensions/exceptions/UnableToBuildPatternMatchException.scala

Wmaarts: Yeah sadly `FailureReason` isn't throwable. You have to return it in an Either monad which makes the whole thing quite complex. 😅 
I agree. So I changed it to a custom exception now. 👍

legopiraat:
Liking it thanks for the help 👍

PR #107, opened 2018-11-23
Add versioning document

### LINKED ISSUE #86: Document versioning strategy
### Summary

To be able to have clear versioning for Stryker4s we need to define and document a strategy.

**Do we want to publish multiple artifacts on the same version?**

1. Publish the same version for all artifacts.
    For example the sbt plugin + the stryker api both on the same version.
    Libraries like Akka are doing this strategy. 
2. Publish each artifact on a different version.
    Current Stryker is doing this. Gives us more flexibility but also more complexity.

legopiraat:
It's a bit minimal at this moment. Not sure what to putt in extra at this moment

Fixes #86

### DIFF
--- a/docs/VERSIONING.md
+++ b/docs/VERSIONING.md
@@ -1,0 +1,15 @@
+# Versioning strategy
+
+This document describes how versioning is managed for the Stryker4s project.
+
+## The strategy
+
+We choose to have one version for the complete project. The main reason for this is because Stryker4s
+will mainly be used as a plugin for your specific used build tool. When using one specific version
+we will be able to make changes in the core module and release everything at once. Also, this makes it
+easier for us to avoid backward compatibility issues which would be very hard to test.
+
+## Versioning
+
+For the actual versioning, we will be applying the standard semver guidelines.
+Don't know the semver guidelines? You can find information about it [here](https://semver.org/).


docs/VERSIONING.md
```diff
@@ -0,0 +1,15 @@
+# Versioning strategy
+
+This document describes how versioning is managed for the Stryker4s project.
+
+## The strategy
+
+We choose to have one version for the complete project. The main reason for this is because Stryker4s
+will mainly be used as a plugin for your specific used build tool. When using one specific version
+we will be able to make changes in the core module and release everything at once. Also, this makes it
```
hugo-vrijswijk:
```suggestion
we will be able to make changes in the core module and release everything at once. This also makes it
```

docs/VERSIONING.md
```diff
@@ -0,0 +1,15 @@
...
+easier for us to avoid backward compatibility issues which would be very hard to test.
```
hugo-vrijswijk:
```suggestion
easier for us to avoid backward compatibility and versions conflict issues which would be very hard to test.
```

legopiraat:
@hugo-vrijswijk updated. i guess this can be merged now?

PR #109, opened 2018-11-27
Configurable mutation score threshold

Wmaarts:
### In stryker4s.conf
```
  thresholds {
    break=10.0
  }
```

### Considerations
- Not sure if throwing an exception or using `sys.exit(exitcode)` is better practice.
   - Throwing exception gives an irrelevant stacktrace in console
   - `sys.exit(exitcode)` possibly has unwanted side effects I'm not sure about


core/src/main/scala/stryker4s/run/report/Reporter.scala
```diff
@@ -1,13 +1,27 @@
...
+class Reporter extends Logging{
 
   /**
     * Generate a report for each reporter that is available.
     */
   def report(runResult: MutantRunResults)(implicit config: Config): Unit = {
     config.reporters.foreach { _.report(runResult) }
   }
+
+  def determineExitCode(runResults: MutantRunResults)(implicit config: Config): Int = {
```
hugo-vrijswijk: Not sure if determining an exit code really is the responsibility of a reporter. I think I'd prefer to have a separate class for it

core/src/main/scala/stryker4s/config/Thresholds.scala
```diff
@@ -0,0 +1,6 @@
+package stryker4s.config
+
+case class Thresholds(// high: Double = 80, // TODO: Not supported yet
+                      // low: Double = 60,  // TODO: Not supported yet
+                      break: Option[Double] = None) {
```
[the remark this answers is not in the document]
Wmaarts: I did want to support at least 2 digits after the comma initially, but thinking about it now it's probably not worth the effort.

Wmaarts:
Ironic how I forgot to write tests. That's fixed now. 😅 
Other feedback from @hugo-vrijswijk has also been implemented.

I was thinking about adding some validation, since you could set the value of 'break' to any number now (should be 0-100). 

Also I didn't implement 'high' and 'low' yet ([as it is in strykerJs](https://github.com/stryker-mutator/stryker/blob/master/packages/stryker/README.md#thresholds-object)) since I'm not sure what to do with these numbers (there's no colored logging yet in Stryker4s :p).

Any thoughts/suggestions on this? 😄

hugo-vrijswijk:
Hmm, perhaps we could configure this as logging info/warn/error. With `break` as the optional value (lets do Option, not null like StrykerJS haha). And later on improve it with colours. 

Running mutation tests on the ThresholdChecker also shows some untested boundaries that would be good to test:

```log
Undetected mutants:
core\src\main\scala\stryker4s\run\ThresholdChecker.scala:15:57:
	from < to <=

core\src\main\scala\stryker4s\run\ThresholdChecker.scala:18:57:
	from >= to >
```

Also, I noticed that both locally and on the build server not all tests are actually run. The `Stryker4sTest` calls `sys.exit(0)` which exits the tests with exit code 0 (so, green build! Yay!). Maybe we should change some of the logic of the `Stryker4s` class, or maybe the test (as it tests quite a lot at the moment and is more of an integration test).

core/src/main/scala/stryker4s/config/Config.scala
```diff
@@ -12,7 +12,8 @@ case class Config(mutate: Seq[String] = Seq("**/main/scala/**/*.scala"),
                   reporters: List[MutantRunReporter] = List(new ConsoleReporter),
                   logLevel: Level = Level.INFO,
                   files: Option[Seq[String]] = None,
-                  excludedMutations: Exclusions = Exclusions(Set.empty)) {
+                  excludedMutations: Exclusions = Exclusions(Set.empty),
+                  thresholds: Option[Thresholds] = None) {
```
legopiraat: Why is this an option? This could also just be a `TreshHolds()` and putt the default break in the case class to 0.

core/src/main/scala/stryker4s/run/ThresholdChecker.scala
```diff
@@ -0,0 +1,23 @@
...
+
+import grizzled.slf4j.Logging
+import stryker4s.config.Config
+import stryker4s.model.MutantRunResults
+
+class ThresholdChecker extends Logging {
+
+  def determineExitCode(runResults: MutantRunResults)(implicit config: Config): Int = {
+    config.thresholds match {
+      case None =>
```
legopiraat: Based on the previous comment we can also then remove this. Am not a fan of this logging statement, to begin with. This is because of the log statements kinda implying you need to specify this.

core/src/main/scala/stryker4s/run/ThresholdChecker.scala
```diff
@@ -0,0 +1,23 @@
...
+        debug("Threshold not configured. Won\'t fail the build no matter how low your mutation score is.")
```
hugo-vrijswijk: Doesn't the String still compile without the \? I think now it also prints a \

core/src/main/scala/stryker4s/run/ThresholdChecker.scala
```diff
@@ -0,0 +1,23 @@
...
+        debug("Set `thresholds.break` to change this behavior.")
+        0
+      case Some(thresholds) if runResults.mutationScore < thresholds.break =>
+        error(s"Mutation score below threshold! Score: ${runResults.mutationScore}. Threshold: ${thresholds.break}")
+        1
+      case Some(thresholds) if runResults.mutationScore >= thresholds.break =>
+        info(s"Mutation score ${runResults.mutationScore} was above or equal to the configured threshold. ")
```
legopiraat: In my opinion, this log message doesn't add any value, since the mutation score was just logged. and in general, if it didn't break I didn't care.

core/src/main/scala/stryker4s/run/ThresholdChecker.scala
Wmaarts: Actually, apparently both work: with and without the `\`. I'll remove it though.

core/src/main/scala/stryker4s/run/ThresholdChecker.scala
Wmaarts: [I actually copied this logging from stryker](https://github.com/stryker-mutator/stryker/blob/master/packages/stryker/src/ScoreResultCalculator.ts#L30).

I could rewrite it to something that doesn't imply that. But since this is logged as DEBUG anyway I thought it could be useful. 
(Could rewrite it to `Set 'threshold.break' if you want to change this behavior`.)

core/src/test/scala/stryker4s/config/ConfigTest.scala
```diff
@@ -26,6 +26,11 @@ class ConfigTest extends Stryker4sSuite {
            |    command=sbt
            |    type=commandrunner
            |}
+           |thresholds {
+           |    break=0
+           |    high=80
+           |    low=60
+           |}
```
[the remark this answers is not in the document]
hugo-vrijswijk: Yes this is what I meant

core/src/main/scala/stryker4s/config/Thresholds.scala
```diff
@@ -1,6 +1,17 @@
 package stryker4s.config
+import stryker4s.extensions.exceptions.InvalidThresholdValueException
 
-case class Thresholds(// high: Int = 80, // TODO: Not supported yet
-                      // low: Int = 60,  // TODO: Not supported yet
-                      break: Int) {
+case class Thresholds(high: Int = 80, low: Int = 60, break: Int = 0){
+  failIfNotPercentage(high, low, break)
+  if(high <= low)
```
hugo-vrijswijk: Not sure how Stryker does it. But I'm unsure about these being `<=`. I think it's fairly intuitive to put both at the same, if you don't care about the lower end of the two.

core/src/test/scala/stryker4s/run/threshold/ThresholdCheckerTest.scala
```diff
@@ -0,0 +1,122 @@
...
+import stryker4s.scalatest.LogMatchers
+
+import scala.concurrent.duration._
+import scala.language.postfixOps
+
+class ThresholdCheckerTest extends Stryker4sSuite with LogMatchers {
+
+  describe("thresholds") {
+    def interceptThresholdException(high: Int = 80, low: Int = 60, break: Int = 0): Unit = {
+      intercept[InvalidThresholdValueException] {
```
hugo-vrijswijk: [Using exception matchers is more in line with the rest of the tests](http://www.scalatest.org/user_guide/using_matchers#expectedExceptions)

core/src/main/scala/stryker4s/config/Thresholds.scala
Wmaarts: You're correct, it shouldn't be `<=`. Stryker does allow 'low' to equal 'high'.

PR #110, opened 2018-11-28
Fix logging incorrect percentages when mutants are excluded

Wmaarts:
Example of the incorrect logging that happened.
```
INFO  Mutator: 146 Mutant(s) generated.
INFO  Mutator: Of which 142 Mutant(s) are excluded.
INFO  ProcessMutantRunner: Starting test-run 26...
INFO  ProcessMutantRunner: Finished mutation run 26/4 (650%)
INFO  ProcessMutantRunner: Starting test-run 31...
INFO  ProcessMutantRunner: Finished mutation run 31/4 (775%)
```

### DIFF
--- a/core/src/main/scala/stryker4s/mutants/findmutants/MutantMatcher.scala
+++ b/core/src/main/scala/stryker4s/mutants/findmutants/MutantMatcher.scala
@@ -1,22 +1,23 @@
 package stryker4s.mutants.findmutants
 
+import stryker4s.config.Config
 import stryker4s.extensions.mutationtypes._
 import stryker4s.model.Mutant
 
 import scala.meta.{Term, Tree}
 
-class MutantMatcher {
+class MutantMatcher()(implicit config: Config) {
 
   private[this] val stream = Iterator.from(0)
 
-  def allMatchers(): PartialFunction[Tree, Seq[Mutant]] =
+  def allMatchers(): PartialFunction[Tree, Seq[Option[Mutant]]] =
     matchBinaryOperators() orElse
       matchBooleanSubstitutions() orElse
       matchLogicalOperators() orElse
       matchStringMutators() orElse
       matchMethodMutators()
 
-  def matchBinaryOperators(): PartialFunction[Tree, Seq[Mutant]] = {
+  def matchBinaryOperators(): PartialFunction[Tree, Seq[Option[Mutant]]] = {
     case GreaterThanEqualTo(orig) => orig ~~> (GreaterThan, LesserThan, EqualTo)
     case GreaterThan(orig)        => orig ~~> (GreaterThanEqualTo, LesserThan, EqualTo)
     case LesserThanEqualTo(orig)  => orig ~~> (LesserThan, GreaterThanEqualTo, EqualTo)
@@ -25,23 +26,23 @@ class MutantMatcher {
     case NotEqualTo(orig)         => orig ~~> EqualTo
   }
 
-  def matchBooleanSubstitutions(): PartialFunction[Tree, Seq[Mutant]] = {
+  def matchBooleanSubstitutions(): PartialFunction[Tree, Seq[Option[Mutant]]] = {
     case True(orig)  => orig ~~> False
     case False(orig) => orig ~~> True
   }
 
-  def matchLogicalOperators(): PartialFunction[Tree, Seq[Mutant]] = {
+  def matchLogicalOperators(): PartialFunction[Tree, Seq[Option[Mutant]]] = {
     case And(orig) => orig ~~> Or
     case Or(orig)  => orig ~~> And
   }
 
-  def matchStringMutators(): PartialFunction[Tree, Seq[Mutant]] = {
+  def matchStringMutators(): PartialFunction[Tree, Seq[Option[Mutant]]] = {
     case EmptyString(orig)         => orig ~~> StrykerWasHereString
     case NonEmptyString(orig)      => orig ~~> EmptyString
     case StringInterpolation(orig) => orig ~~> EmptyStringInterpolation
   }
 
-  def matchMethodMutators(): PartialFunction[Tree, Seq[Mutant]] = {
+  def matchMethodMutators(): PartialFunction[Tree, Seq[Option[Mutant]]] = {
     case Filter(orig, f)      => orig ~~> (FilterNot, f)
     case FilterNot(orig, f)   => orig ~~> (Filter, f)
     case Exists(orig, f)      => orig ~~> (ForAll, f)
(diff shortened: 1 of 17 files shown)


core/src/main/scala/stryker4s/run/ProcessMutantRunner.scala
```diff
@@ -65,10 +65,9 @@ class ProcessMutantRunner(command: Command, process: ProcessRunner)(implicit con
     MutantRunResults(runResults, calculateMutationScore(totalMutants, detected.size), duration)
   }
 
-  private[this] def runMutant(mutant: Mutant, workingDir: File, subPath: Path): MutantRunResult = {
-    val id = mutant.id
-    info(s"Starting test-run ${id + 1}...")
-    process(command, workingDir, ("ACTIVE_MUTATION", id.toString)) match {
+  private[this] def runMutant(mutant: Mutant, workingDir: File, subPath: Path, id: Int): MutantRunResult = {
+    info(s"Starting test-run $id...")
+    process(command, workingDir, ("ACTIVE_MUTATION", (id - 1).toString)) match {
```
hugo-vrijswijk: Does this work with the correct mutation? The mutation is made based on the mutant id, not the index that's given here.

So if 4 mutants are generated with ids 0, 1, 2, 3, and 2 is excluded, then the `ACTIVE_MUTATION`s set here would be 0, 1, 2

core/src/main/scala/stryker4s/run/ProcessMutantRunner.scala
```diff
@@ -52,9 +52,9 @@ class ProcessMutantRunner(command: Command, process: ProcessRunner)(implicit con
       mutatedFile <- mutatedFiles
       subPath = mutatedFile.fileOrigin.relativePath
       mutant <- mutatedFile.mutants
+      id = mutatedFiles.flatMap(_.mutants).toSeq.indexOf(mutant) + 1
```
hugo-vrijswijk: This search is O(n^2), right? For every mutant the entire list is searched. Something like `zipWithIndex` might be better

legopiraat: Agreed zipWithIndex would be a lot better. We could simply do something like: `(mutatedFile, index) <- mutatedFiles.zipWithIndex`. After that change the log statement to use the index and not touch the current Id i think hugo is right and that it will not run the proper mutations otherwise.

core/src/main/scala/stryker4s/run/ProcessMutantRunner.scala
```diff
@@ -65,9 +65,8 @@ class ProcessMutantRunner(command: Command, process: ProcessRunner)(implicit con
     MutantRunResults(runResults, calculateMutationScore(totalMutants, detected.size), duration)
   }
 
-  private[this] def runMutant(mutant: Mutant, workingDir: File, subPath: Path): MutantRunResult = {
-    val id = mutant.id
-    info(s"Starting test-run ${id + 1}...")
+  private[this] def runMutant(mutant: Mutant, workingDir: File, subPath: Path, id: Int): MutantRunResult = {
+    info(s"Starting test-run $id...")
     process(command, workingDir, ("ACTIVE_MUTATION", id.toString)) match {
```
legopiraat: You could pass in the index of the mutation so the log statement is correct. but you will still have to use mutant.id to run the process otherwise you will have the wrong active mutant.

core/src/main/scala/stryker4s/run/ProcessMutantRunner.scala
```diff
@@ -48,14 +48,16 @@ class ProcessMutantRunner(command: Command, process: ProcessRunner)(implicit con
...
-    val runResults = for {
+    val mutantsWithPath = for {
       mutatedFile <- mutatedFiles
       subPath = mutatedFile.fileOrigin.relativePath
       mutant <- mutatedFile.mutants
-      id = mutatedFiles.flatMap(_.mutants).toSeq.indexOf(mutant) + 1
-    } yield {
+    } yield (mutant, subPath)
+
+    val runResults = mutantsWithPath.zipWithIndex.map{ case ((mutant, subPath), index) =>
```
Wmaarts: I couldn't get it to work within the for-comprehension, so this was the cleanest rewrite I could think of.

legopiraat: You should be able to do this:
```scala
val mutantsWithPath = for {
      (mutatedFile, index) <- mutatedFiles.zipWithIndex
      subPath = mutatedFile.fileOrigin.relativePath
      mutant <- mutatedFile.mutants
    } yield {
```

hugo-vrijswijk: Wouldn't that give an index per file instead of per mutation?

Wmaarts: But this generates an index for each file, but we need an index for each mutant. With this code all mutants from the same file get the same id.

legopiraat: Oow right didn't think of that.

hugo-vrijswijk: A better fix might be to assign IDs before excluding mutations

Wmaarts: Right now the mutants are first created in MutantMatcher.scala and then excluded in MutantFinder.scala. So excluding logic would have to be built in MutantMatcher.scala. But I'm not sure if that's clean. 

Implicitly passing the Config to MutantMatchers works for example. Then we can do the exclusion logic there. Example:
```
def ~~>(mutated: MethodMutator, f: String => Term)(implicit config: Config): Seq[Mutant] = {
  if (config.excludedMutations.exclusions.contains(mutated.mutationName))
    Mutant(-1, original, mutated.apply(f), mutated) :: Nil // Excluded
  else
    Mutant(stream.next, original, mutated.apply(f), mutated) :: Nil
}
```

This gives the included mutators correct indexes: 0, 1, 2, 3, etc. and all excluded -1. (obviously -1 is a bad idea but I was just trying it out)

This feels like logic that shouldn't be in this class. But I could be wrong. What do you think @hugo-vrijswijk ?

hugo-vrijswijk: I'm not sure. The MutantMatcher taking user configuration into consideration for what mutants to match on makes sense to me

core/src/main/scala/stryker4s/mutants/findmutants/MutantMatcher.scala
```diff
@@ -1,22 +1,23 @@
...
 
 import scala.meta.{Term, Tree}
 
-class MutantMatcher {
+class MutantMatcher()(implicit config: Config) {
 
   private[this] val stream = Iterator.from(0)
 
-  def allMatchers(): PartialFunction[Tree, Seq[Mutant]] =
+  def allMatchers(): PartialFunction[Tree, Seq[Option[Mutant]]] =
```
hugo-vrijswijk: Instead of making a `Seq[Option[Mutant]]`, why not still return a `Seq[Mutant]`, but just filter out the mutants that are excluded. So sometimes giving an empty list

Issue #95, opened 2018-10-18
String and Boolean mutation inside Annotations


scelestino:
### Summary

Strings and Booleans inside Annotations should be escaped like `case`.

```
case class Foo(bar: Option[String],
                       @JsonProperty("type") bazType: Baz)
```

```
annotation argument needs to be a constant; found: scala.sys.`package`.props.get("ACTIVE_MUTATION").orElse[String](scala.sys.`package`.env.get("ACTIVE_MUTATION")) match {
```

### Stryker config

```
stryker4s {
  mutate: ["app/**/*.scala"],
  base-dir: "/home/scelestino/dev/myapp"
}
```

### Stryker4s environment

The SBT branch.

```
stryker4s 0.1.0-SNAPSHOT
```

### Your Environment

Scala 2.12.7
SBT 1.0.3
OS Ubuntu

scelestino:
I think the same. Maybe we can skip the tree if it is an annotation.

I'm going to check that.

scelestino:
There is one point where we can remove this kind of trees from the mutation list.

Currently `MutantFinder.findMutants` uses a standard `collect` that traverse the entire Tree (every node and leaf). As you did with `ImplicitTreeExtensions`, we can create a custom `Traverser` that only visit valid nodes.

Here you are `collect` that doesn't lookup inside Annot nodes (tree for annotations)
```
  val liftedFn = fn.lift
  val buf = scala.collection.mutable.ListBuffer[Unit]()

  object traverser extends Traverser {
    override def apply(tree: Tree): Unit = tree match {
      case Annot(_) => // nothing
      case _ =>
        liftedFn(tree).foreach(buf += _)
        super.apply(tree)
    }
  }
  traverser(exampleTree)
```

Even we can do the same with `Case(pattern, cond, body)` to skip `pattern` (I know that issue is solved, but this is another way to do it)

hugo-vrijswijk:
Funnily enough, I actually wrote a test for an annotation today. Though that was also the first time I did so. 

But don't annotations require constant values? That'd mean we can't put a pattern match in it

scelestino:
> But don't annotations require constant values? That'd mean we can't put a pattern match in it

Exactly, that's why my proposal is skip Annot Tree from MutantFinder.findMutants

scelestino:
@hugo-vrijswijk can you upload that test so I can try to fix it?

hugo-vrijswijk:
[...]

PR #112, opened 2018-12-04
Skip mutations for Annotations

Wmaarts:
#### What is does
Before creating a Mutant, this checks if the Term isn't inside of an annotation. If it is, skip creating the mutant entirely (it doesn't log this).
(In #95 there's talk about String and Boolean, but this implementation doesn't check for that: it simply skips everything inside annotations.)

#### How it works
Recursively goes up the tree until it finds a Mod.Annot (or until it cannot go further up).

It does not use a custom `Traverser` as suggested in #95. That implementation could be more efficient, but this implementation isn't exactly performance intensive either.

### DIFF
--- a/core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
+++ b/core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
@@ -67,7 +67,6 @@ class FileCollector(implicit config: Config) extends SourceCollector with Loggin
   private[this] def listAllFiles(): Option[Iterable[File]] = {
     Option(config.baseDir
       .listRecursively
-      .filterNot(file => file.isDirectory)
       .filterNot(file => file.pathAsString.contains(s"${pathSeparator}target$pathSeparator"))
       .toIterable)
   }
--- a/core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
+++ b/core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
@@ -67,6 +67,7 @@ class FileCollector(implicit config: Config) extends SourceCollector with Loggin
   private[this] def listAllFiles(): Option[Iterable[File]] = {
     Option(config.baseDir
       .listRecursively
+      .filterNot(file => file.isDirectory)
       .filterNot(file => file.pathAsString.contains(s"${pathSeparator}target$pathSeparator"))
       .toIterable)
   }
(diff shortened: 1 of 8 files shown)


util/src/main/scala/stryker4s/extensions/TreeExtensions.scala
```diff
@@ -80,5 +80,17 @@ object TreeExtensions {
       override def apply(tree: Tree): Tree =
         liftedFn(tree).getOrElse(super.apply(tree))
     }
+
+    /** Returns if a tree is contained in an annotation.
+      * Recursively going up the tree until a annotation is found.
```
hugo-vrijswijk:
```suggestion
      * Recursively going up the tree until an annotation is found.
```

Typo

core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
```diff
@@ -67,6 +67,7 @@ class FileCollector(implicit config: Config) extends SourceCollector with Loggin
   private[this] def listAllFiles(): Option[Iterable[File]] = {
     Option(config.baseDir
       .listRecursively
+      .filterNot(file => file.isDirectory)
```
Wmaarts: Oh woops, didn't mean for this to be here.

This line of code fixes a problem I had for my test project. Because I did not use a git repo the FileCollector chose to copy the whole directory. But the directories are part of this list too. 

So then what happens is an empty file '.idea' is created (which should be a directory). And then it tries to created a directory '.idea' to put a file in and I get `java.nio.file.FileAlreadyExistsException: C:\path\to\stryker\target\stryker4s-6944695956008966908\.idea`.

core/src/test/scala/stryker4s/mutants/applymutants/MatchBuilderTest.scala
```diff
@@ -154,6 +154,7 @@ class MatchBuilderTest extends Stryker4sSuite with TreeEquality {
     val topStatement = source.find(origStatement).value.topStatement()
     val mutant = mutants
       .map(m => topStatement transformOnce { case orig if orig.isEqual(origStatement) => m } get)
+      .filterNot(_.isInAnnotation)
```
hugo-vrijswijk: If annotations are already excluded in the `MutantMatcher`, is this still needed?

Wmaarts: Good point, it's actually not needed.

core/src/main/scala/stryker4s/mutants/findmutants/FileCollector.scala
hugo-vrijswijk: This is already fixed in the `sbt-plugin` branch. Could you revert it to prevent a merge conflict? 😀

Issue #117, opened 2018-12-21
Integration tests


hugo-vrijswijk:
We should have some integration tests. Some with `sbt-scripted` to test the sbt plugin

legopiraat:
Am working on this a little bit, will continue soon

PR #128, opened 2019-01-18
Re organize project modules

### LINKED ISSUE #120: Re-organize project modules
I think we should re-organize the submodule structure. Right now there's no real functional difference between the `util` and `core` packages. We could combine those two into one. And while we're at it create a new `command-runner` package that contains all code needed for the command-runner, so that's not needlessly added into packages that don't need it (like the sbt plugin).

This way we would end up with 3 packages:
- sbt-stryker4s
- stryker4s-core
- stryker4s-command-runner (or something similar)

hugo-vrijswijk:
#### What it does

Reorganized the project modules into the following structure:

root (`stryker4s`)
--core (`stryker4s-core`)
--runners (not a module, just a folder)
----command-runner (`stryker4s-command-runner`)
----sbt (`sbt-stryker4s`)

Once we have integration tests (see #117), we can add it in a seperate module `integration-tests`

#### How it works

Sbt magic™

Separating the command-runner into a separate module also means we can have just an api dependency on log4j for the core project, with whatever the runner is (sbt, command-runner) providing the implementation. The command-runner now has a dependency on `log4j-slf4j-impl` that the core project had before.


legopiraat:
Changed a little bit in the lib configuration. Now it was included in all modules, but as a best practice online include what you need. So thats what i changed with my last commit and i think its ready for merge.

hugo-vrijswijk:
Looks good to me. Merge when you want 👍

PR #129, opened 2019-01-21
Test against multiple JDK's on Travis

hugo-vrijswijk:
I've added multiple JDK versions to Travis. As now we only test against 1 version on Travis. This should hopefully prevent any problems with different JDK's. As we only use Scala libraries, we probably won't have many problems, but it's still good to test.

I've also added a pre-release version, for us to keep an eye on any problems future JDKs might cause.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,17 @@ sudo: false
 
 language: scala
 
+jdk:
+  - oraclejdk8
+  - oraclejdk11
+  - openjdk8
+  - openjdk11
+  - openjdk-ea
+
+matrix:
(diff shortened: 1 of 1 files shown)


hugo-vrijswijk:
So, 5 JDK's is quite a lot in my opinion, and maybe a bit overkill, as Scala shouldn't be too dependent on different JDK's. @stryker-mutator/stryker4s, how would you feel about the following list:

- oraclejdk8
- openjdk11
- openjdk-ea

hugo-vrijswijk:
Removed EA as discussed IRL with @legopiraat.

PR #130, opened 2019-01-22
Fix pattern match mutations

### LINKED ISSUE #124: Mutations in a pattern match with alternatives
If a pattern match has alternatives (`Pat.Alternative` in Scalameta AST) the tree isn't mutated correctly.

The following code:

```scala
foo match {
  case "string1" | "string2" => bar
}
```

Gives a `UnableToBuildPatternMatchException` with the following message: `Invalid transformation of Pat.Alternative.rhs: Lit.String -> Term.Match. From: Lit.String("convertTo"), to: Term.Match(Term.Apply(Term.Select(Term.Apply(Term.Select(Term.Select(Term.Name("sys"), Term.Name("props")), Term.Name("get")), List(Lit.String("ACTIVE_MUTATION"))), Term.Name("orElse")), List(Term.Apply(Term.Select(Term.Select(Term.Name("sys"), Term.Name("env")), Term.Name("get")), List(Lit.String("ACTIVE_MUTATION"))))), List(Case(Pat.Extract(Term.Name("Some"), List(Lit.String("35"))), None, Lit.String("")), Case(Pat.Wildcard(), None, Lit.String("convertTo"))))`.

The same happens when a `Pat.Extract` is in the pattern match (`case "convertFrom" / "convertTo" => bar`). A solution should probably aim at fixing both. I've also saved an example [here](https://astexplorer.net/#/gist/504abb270f23822d53b6a9a4f61e2c71/523ec64d2b8c104f2338329eb25121d308a7a62b) so anyone can look at the AST.

hugo-vrijswijk:
#### What it does

This fixes the issue of code being generated that is not compileable. E.g. `case "user" / "create" => createUser(foo)` would cause code generation that could not compile.

#### How it works

Instead of going up a tree once to check if a statement is inside a pattern match, it checks if a statement is inside a pattern match, if so, it finds that pattern match and goes up until a `Term` is found. That means the statement is safe to build.


legopiraat:
@hugo-vrijswijk 
Did you also test this against the hello world from http4s? because that didnt work either and this should fix it i believe.

hugo-vrijswijk:
Yes, I've tried on both a reference project of my own and the [http4s.g8](https://github.com/http4s/http4s.g8). Both worked perfectly 👍.

core/src/main/scala/stryker4s/extension/TreeExtensions.scala
```diff
@@ -28,25 +29,36 @@ object TreeExtensions {
...
+      private def parentOf(tree: Tree): Option[Term] = tree.parent match {
+        case Some(caseTree: Case) => findInsideCase(caseTree) // Case is found, go into that
+        case Some(other)          => parentOf(other) // Keep going up
+        case None                 => None // Top of tree is reached, we better stop
+      }
+
+      @tailrec
+      private def findInsideCase(caseTree: Tree): Option[Term] = caseTree.parent match {
+        case Some(term: Term) => Some(term) // Great success!
+        case Some(other)      => findInsideCase(other) // Keep going up
```
Wmaarts: `topStatement()` doesn't have comments explaining what each case does. I like keeping comments for clarity though, but consistency might be a thing? 
Also, `// Great success!` might not be very descriptive. ^^

hugo-vrijswijk: Fixed. And much more...

PR #131, opened 2019-01-23
Automated release

### LINKED ISSUE #118: Automated release pipeline
It'd be great to have an automated release pipeline. [sbt-release](https://github.com/sbt/sbt-release) seems like a good tool to use. 

EDIT: [`sbt-ci-release`](https://github.com/olafurpg/sbt-ci-release) also seems to do a lot of what we want and is made for Travis.

Ideas:
- Do we want a commit for each release, similar to the maven release plugin, updating the version number (though I think that's not used much anymore)?
- Could we have automated release notes based on merged PR's? We'd have to standardize the PR format
- Automatic pushing to sonatype
- Automatic tag push (for github releases)

hugo-vrijswijk:
### Fixes #118 

As there were a couple changes in the Travis.yml with #129, I've used that as a base. That PR should be merged before this one, to get a better diff.

#### What it does

Uses the sbt-ci-release plugin for automated releases from Travis. Every commit on master is now built and uploaded as a snapshot, and tags are built to releases.

#### How it works

Using the sbt-ci-release plugin, I've added another stage to the Travis build. After `Test` is successful, `Release` will start and build, sign and upload a release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,17 +3,27 @@ language: scala
 scala:
   - 2.12.8
 
-jdk:
-  - oraclejdk8
-  - openjdk11
-  - openjdk-ea
+stages:
+  - name: test
+  - name: release
+    if: (branch = master AND type = push) OR (tag IS present)
+
+jobs:
+  include:
+    - stage: test
+      script: sbt clean coverage test coverageReport
+      jdk:
+        - oraclejdk8
+        - openjdk11
+        - openjdk-ea
+    # run ci-release only if previous stages passed
+    - stage: release
+      script: sbt ci-release
+      jdk: openjdk11
 
 matrix:
   allow_failures:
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -6,9 +6,21 @@ scala:
...
   - oraclejdk8
   - openjdk11
-  
-script:
-  - sbt clean coverage test coverageReport
+
+stages:
+  - name: test
+  - name: release
+    if: (branch = master AND type = push) OR (tag IS present)
```
legopiraat: How does this plugin know how to update the version number? or will this override the current published snapshot version?

hugo-vrijswijk:
Good idea! I'll add it

.travis.yml
legopiraat: Oke good to know!

PR #136, opened 2019-01-24
Copy jvm systemproperties to forked test JVMs

### LINKED ISSUE #119: Running initial test run doesn't copy JVM system settings
### Summary

Running initial test run doesn't copy JVM system settings.

My tests require a JVM system setting (`-Dconfig.file=...`) to point to a configuration file for the environment where the tests are ran. Running `sbt -Dconfig.file=... test` runs fine. Running `sbt -Dconfig.file=... "project app" stryker` (since I'm using a multi-module structure) fails with:

    [info] *** 4 SUITES ABORTED ***
    [info] *** 14 TESTS FAILED ***
    [error] Failed tests:
    [error] 	hyperion.MeterAgentSpec
    [error] 	hyperion.DailyHistoryActorSpec
    [error] 	hyperion.RecentHistoryActorSpec
    [error] Error during tests:
    [error] 	hyperion.database.DatabaseActorSpec
    [error] 	hyperion.AppSettingsSpec
    [error] 	hyperion.rest.AppInfoServiceSpec
    [error] (Test / test) sbt.TestsFailedException: Tests unsuccessful
    [error] stryker4s.extensions.exceptions.InitialTestRunFailedException: Initial test run failed. Please make sure your tests pass before running Stryker4s.
    [error] stryker4s.extensions.exceptions.InitialTestRunFailedException: Initial test run failed. Please make sure your tests pass before running Stryker4s.
    [error] Use 'last' for the full log.

See [the build log](https://circleci.com/gh/mthmulders/hyperion/348) and [the source code repository](https://github.com/mthmulders/hyperion/tree/mutation-testing) for more details.

### Stryker config

```hocon
stryker4s {
  base-dir = "app"
  log-level = debug
  mutate = [
    "app/src/main/scala/**/*.scala"
  ]
  reporters = [
    console
  ]
}
```

### Stryker4s environment

```
stryker4s 0.1.0
```

### Your Environment
| Scala version        | 2.12.8 |
| Build tool & version | SBT 1.2.8 | 
| Operating System     | MacOS Mojave (10.14.2) |

Wmaarts:
#### What it does
- Introduces a config option in `stryker4s.conf` that will make sure the included system properties are copied.

#### How it works
Retrieves all system properties by calling `sys.props`, then filters those that were defined in config and copies these to the `javaOptions` TaskKey of sbt.

#### Also includes
I had to refactor the way `excluded-mutations` are parsed, because it used to parse and validate every `List[String]` in the config, which includes the `system-properties` config field I introduced. 

So I created `case class ExcludedMutations` which solves the issue.

### DIFF
--- a/core/src/main/scala/stryker4s/config/Config.scala
+++ b/core/src/main/scala/stryker4s/config/Config.scala
@@ -12,8 +12,7 @@ case class Config(mutate: Seq[String] = Seq("**/main/scala/**/*.scala"),
                   logLevel: Level = Level.INFO,
                   files: Option[Seq[String]] = None,
                   excludedMutations: ExcludedMutations = ExcludedMutations(),
-                  thresholds: Thresholds = Thresholds(),
-                  systemProperties: Set[String] = Set.empty) {
+                  thresholds: Thresholds = Thresholds()) {
 
   def toHoconString: String = {
     import stryker4s.config.implicits.ConfigWriterImplicits._
--- a/core/src/main/scala/stryker4s/config/Config.scala
+++ b/core/src/main/scala/stryker4s/config/Config.scala
@@ -11,8 +11,9 @@ case class Config(mutate: Seq[String] = Seq("**/main/scala/**/*.scala"),
                   reporters: List[MutantRunReporter] = List(new ConsoleReporter),
                   logLevel: Level = Level.INFO,
                   files: Option[Seq[String]] = None,
-                  excludedMutations: Set[String] = Set.empty,
-                  thresholds: Thresholds = Thresholds()) {
+                  excludedMutations: ExcludedMutations = ExcludedMutations(),
+                  thresholds: Thresholds = Thresholds(),
+                  systemProperties: Set[String] = Set.empty) {
 
   def toHoconString: String = {
     import stryker4s.config.implicits.ConfigWriterImplicits._
--- a/runners/sbt/src/main/scala/stryker4s/sbt/SbtMutantRunner.scala
+++ b/runners/sbt/src/main/scala/stryker4s/sbt/SbtMutantRunner.scala
@@ -60,9 +60,8 @@ class SbtMutantRunner(state: State, processRunner: ProcessRunner)(implicit confi
       fork in Test := true,
       javaOptions in Test ++= {
         val props = sys.props.toList
-        // TODO -> Make configurable
-        props.filterNot(s => s._1 == "java.class.path")
-          .map {
+
+        props.filter(s => config.systemProperties.contains(s._1)).map {
           case (key, value) => s"-D$key=$value"
         }
       },
--- a/runners/sbt/src/main/scala/stryker4s/sbt/SbtMutantRunner.scala
+++ b/runners/sbt/src/main/scala/stryker4s/sbt/SbtMutantRunner.scala
@@ -37,6 +37,23 @@ class SbtMutantRunner(state: State, processRunner: ProcessRunner)(implicit confi
     }
   }
 
+  private lazy val filteredSystemProperties = {
+    // Matches strings that start with one of the options between brackets
+    val regex = "^(java|sun|file|user|jna|os|sbt|jline|awt|user).*"
+
+    val filteredProps =
+      sys.props.toList
+        .filterNot { case (key, _) => key.matches(regex) }
+        .map { case (key, value) => s"-D$key=$value" }
+
+    if (filteredProps.nonEmpty) {
+      debug("System properties added to the forked JVM")
+      filteredProps.foreach(debug(_))
+    }
+
+    filteredProps
+  }
+
   private[this] def settings(tmpDir: File): Seq[Def.Setting[_]] = {
     val mainPath = {
       extracted
@@ -59,11 +76,7 @@ class SbtMutantRunner(state: State, processRunner: ProcessRunner)(implicit confi
     Seq(
       fork in Test := true,
       javaOptions in Test ++= {
-        val props = sys.props.toList
-
-        props
-          .filter { case (key, _) => config.systemProperties.contains(key) }
-          .map { case (key, value) => s"-D$key=$value" }
+        filteredSystemProperties
       },
       scalaSource in Compile := tmpDir.toJava / mainPath,
       scalaSource in Test := tmpDir.toJava / testPath
--- a/runners/sbt/src/main/scala/stryker4s/sbt/SbtMutantRunner.scala
+++ b/runners/sbt/src/main/scala/stryker4s/sbt/SbtMutantRunner.scala
@@ -16,30 +16,30 @@ class SbtMutantRunner(state: State, processRunner: ProcessRunner, sourceCollecto
     implicit config: Config)
     extends MutantRunner(processRunner, sourceCollector) {
 
-  private lazy val filteredSystemProperties = {
+  private lazy val filteredSystemProperties: Option[List[String]] = {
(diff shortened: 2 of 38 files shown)


core/src/main/scala/stryker4s/config/Config.scala
```diff
@@ -11,8 +11,9 @@ case class Config(mutate: Seq[String] = Seq("**/main/scala/**/*.scala"),
                   reporters: List[MutantRunReporter] = List(new ConsoleReporter),
                   logLevel: Level = Level.INFO,
                   files: Option[Seq[String]] = None,
-                  excludedMutations: Set[String] = Set.empty,
-                  thresholds: Thresholds = Thresholds()) {
+                  excludedMutations: ExcludedMutations = ExcludedMutations(),
+                  thresholds: Thresholds = Thresholds(),
+                  systemProperties: Set[String] = Set.empty) {
```
hugo-vrijswijk: Shouldn't this be in a case class as well (or a `Set[SomeOtherType]`) to prevent problems should we ever want to add another set?

Wmaarts: Having another `Set` is not an issue in itself. I had to change it because ExcludedMutations has special behaviour: it only allows predefined values (EqualityOperator, BooleanLiteral, LogicalOperator, etc). 

`systemProperties` is just a normal `Set` so I don't think it needs a case class. :)

legopiraat: I don't think a set is the correct list to use. certainly when you are doing this `s"-D$key=$value"` it's most certainly possible to have multiple keys that would have the same value we don't know this. So I would assume duplicates will be very much possible.

So either expect full `["-Dconfig=value"]` or allow duplicates

runners/sbt/src/main/scala/stryker4s/sbt/SbtMutantRunner.scala
```diff
@@ -58,6 +58,13 @@ class SbtMutantRunner(state: State, processRunner: ProcessRunner)(implicit confi
 
     Seq(
       fork in Test := true,
+      javaOptions in Test ++= {
+        val props = sys.props.toList
+
+        props.filter(s => config.systemProperties.contains(s._1)).map {
```
legopiraat: Whats an `s` ? Please use descriptive variables names

legopiraat:
If the user needs to write the whole javaOptions that will be a bit to much. We could also introduce sbt specific config by adding a key, and using it like

build.sbt
```
stryker4sJvmOptions ++= Map("key","value")
```

hugo-vrijswijk:
Hmm, I am not sure. We'd by redefining existing keys, which is [considered bad practice](https://www.scala-sbt.org/1.0/docs/Plugins-Best-Practices.html#Reuse+existing+keys)

core/src/main/scala/stryker4s/config/Config.scala
legopiraat: I came up with another argument this morning.

Some properties are not key value, for instance, booleans are not mandatory to be key value pairs.

Wmaarts:
> So we can use either this PR, which is extra configuration for the user, or change to appendWithSession, which is extra reading.

The main issue I have with the `appendwithSession` implementation right now is how the command looks. :/

`sbt ';set javaOptions in Test += "-Dconfig.value=foo"; stryker'` 
versus
`sbt -Dconfig.value=foo stryker'` 

And what if you use two properties:
`sbt "project submodule" ';set javaOptions in Test ++= Seq("-Dconfig.file=path/to/somesortof/config.conf", "-Dconfig.outputfolder=the/path/to/an/output/folder"); stryker'` 😨 
versus
`sbt -Dconfig.file=path/to/somesortof/config.conf -Dconfig.outputfolder=the/path/to/an/output/folder "project submodule" stryker`

Other than that, `appendWithSession` looks like a good solution to be honest.

core/src/main/scala/stryker4s/config/Config.scala
legopiraat: oooooww yes then we misunderstanded for sure

Wmaarts:
I changed the logic to copy all system properties except for the ones matching this regex:
```
    // Matches strings that start with one of the options between brackets
    val regex = "^(java|sun|file|user|jna|os|sbt|jline|awt|user).*"
```

Regex was created based on [jvm properties](https://docs.oracle.com/javase/tutorial/essential/environment/sysprop.html) and other properties found while testing.

Could use a review again. 👀

runners/sbt/src/main/scala/stryker4s/sbt/SbtMutantRunner.scala
```diff
@@ -16,8 +16,27 @@ class SbtMutantRunner(state: State, processRunner: ProcessRunner, sourceCollecto
...
+    // Matches strings that start with one of the options between brackets
+    val regex = "^(java|sun|file|user|jna|os|sbt|jline|awt|user).*"
+
+    val filteredProps =
+      sys.props.toList
+        .filterNot { case (key, _) => key.matches(regex) }
+        .map { case (key, value) => s"-D$key=$value" }
+
+    if (filteredProps.nonEmpty) {
+      debug("System properties added to the forked JVM")
```
hugo-vrijswijk: Probably cleaner to put this in one debug statement, something like `debug("System properties added to the forked JVM: ${filteredProps.mkString(", ")}")

runners/sbt/src/main/scala/stryker4s/sbt/SbtMutantRunner.scala
```diff
@@ -16,8 +16,27 @@ class SbtMutantRunner(state: State, processRunner: ProcessRunner, sourceCollecto
...
+      filteredProps.foreach(debug(_))
+    }
+    filteredProps
+  }
+
   private val settings: Seq[Def.Setting[_]] = Seq(
     fork in Test := true,
+    javaOptions in Test ++= {
```
hugo-vrijswijk: Should probably only do this if the properties are not empty. To save the performance hit of sbt adding and setting an empty list

runners/sbt/src/main/scala/stryker4s/sbt/SbtMutantRunner.scala
```diff
@@ -37,6 +37,23 @@ class SbtMutantRunner(state: State, processRunner: ProcessRunner)(implicit confi
     }
   }
 
+  private lazy val filteredSystemProperties = {
+    // Matches strings that start with one of the options between brackets
+    val regex = "^(java|sun|file|user|jna|os|sbt|jline|awt|user).*"
```
legopiraat: user is duplicated

runners/sbt/src/main/scala/stryker4s/sbt/SbtMutantRunner.scala
```diff
@@ -16,11 +16,30 @@ class SbtMutantRunner(state: State, processRunner: ProcessRunner, sourceCollecto
...
+
   private val settings: Seq[Def.Setting[_]] = Seq(
     fork in Test := true,
     scalaSource in Compile := tmpDirFor(Compile).value,
     scalaSource in Test := tmpDirFor(Test).value,
-  )
+  ) ++
+    filteredSystemProperties.map(properties => {
+      debug(s"System properties added to the forked JVM: ${properties.mkString(",")}")
+      javaOptions ++= properties
```
hugo-vrijswijk:
```suggestion
      javaOptions in Test ++= properties
```

PR #138, opened 2019-01-24
Refactor sbt plugin

hugo-vrijswijk:
I've refactored the sbt plugin a little bit. The `SbtMutantRunner` is little cleaned up, and I've refactored the main settings (everything but the active mutation) to only be calculated once. Unfortunately, they are still applied for each mutation, instead of once, which would be a far larger performance improvement. I've spent some time trying to get that to work, but couldn't (yet).

I've also changed the main `stryker` command to be a `Task`, instead of a `Command`, as recommended by the [sbt documentation](https://www.scala-sbt.org/1.0/docs/Plugins-Best-Practices.html#Use+settings+and+tasks.+Avoid+commands.)

### DIFF
--- a/runners/sbt/src/main/scala/stryker4s/sbt/SbtMutantRunner.scala
+++ b/runners/sbt/src/main/scala/stryker4s/sbt/SbtMutantRunner.scala
@@ -56,13 +56,13 @@ class SbtMutantRunner(state: State, processRunner: ProcessRunner, sourceCollecto
     javaOptions in Test += s"-DACTIVE_MUTATION=${String.valueOf(mutation)}"
 
   private def tmpDirFor(conf: Configuration): Def.Initialize[JFile] = {
-    val sourceDef = pathOf(scalaSource, conf)
-    val baseDef = pathOf(baseDirectory, conf)
+    val sourceDirDef = (scalaSource in conf)(_.absolutePath)
+    val baseDirDef = (baseDirectory in conf)(_.absolutePath)
 
-    (sourceDef zipWith baseDef)(_ diff _)(tmpDir.toJava / _)
+    sourceDirDef.zipWith(baseDirDef) { (sourceDir, baseDir) =>
+      val relativePath = sourceDir diff baseDir
+
+      tmpDir.toJava / relativePath
+    }
   }
-
-  private def pathOf(key: SettingKey[JFile], conf: sbt.Configuration): Def.Initialize[String] =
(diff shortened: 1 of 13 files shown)


runners/sbt/src/main/scala/stryker4s/sbt/SbtMutantRunner.scala
```diff
@@ -1,73 +1,68 @@
...
-    Seq(
-      fork in Test := true,
-      scalaSource in Compile := tmpDir.toJava / mainPath,
-      scalaSource in Test := tmpDir.toJava / testPath
-    )
+  private def tmpDirFor(conf: Configuration): Def.Initialize[JFile] = {
+    val sourceDef = pathOf(scalaSource, conf)
+    val baseDef = pathOf(baseDirectory, conf)
 
+    (sourceDef zipWith baseDef)(_ diff _)(tmpDir.toJava / _)
```
Wmaarts: I'm having a hard time understanding what exactly happens here. Fancy oneliners are cool but I think this could do with a bit more readability. 😀

runners/sbt/src/main/scala/stryker4s/sbt/SbtMutantRunner.scala
```diff
@@ -1,73 +1,68 @@
...
 
-  private[this] def mutationSetting(mutation: Int): Seq[Def.Setting[_]] = {
-    Seq(
-      // Set active mutation
-      javaOptions in Test += s"-DACTIVE_MUTATION=${String.valueOf(mutation)}"
-    )
-  }
+  private def pathOf(key: SettingKey[JFile], conf: sbt.Configuration): Def.Initialize[String] =
+    (key in conf)(_.absolutePath)
+
```
Wmaarts: I'm on the fence about this. If you repeat a small piece of code twice, do you immediately need to create a function for it? :')

hugo-vrijswijk: Hmm, I agree. It was longer when I first wrote it, but I refactored it to be shorter. But I will inline it and try to clear it up a bit

runners/sbt/src/main/scala/stryker4s/sbt/SbtMutantRunner.scala
Wmaarts: Yeah this is better. Fewer underscores helps. :)

PR #140, opened 2019-01-28
Log a warning when 0 mutants are found

### LINKED ISSUE #132: Log warning when 0 mutants are found
When zero mutants are found, something is likely wrong. We should log a warning about this instead of just giving a 0.0% mutation score

hugo-vrijswijk:
#### What it does

Logs when 0 mutants are found, letting the user know something could be wrong with their configuration.

#### How it works

Logging magic 🔮

### DIFF
--- a/core/src/main/scala/stryker4s/mutants/Mutator.scala
+++ b/core/src/main/scala/stryker4s/mutants/Mutator.scala
@@ -51,10 +51,9 @@ class Mutator(mutantFinder: MutantFinder, transformer: StatementTransformer, mat
     if (excludedMutants > 0) {
       info(s"Of which $excludedMutants Mutant(s) are excluded.")
     }
-    if (totalMutants == 0) {
-      warn("It's a mutant-free world, nothing to test.")
-      warn("This is likely an issue with configuration.")
-      warn("Please check your configuration and try again.")
+    if (includedMutants == 0) {
+      info("It's a mutant-free world, nothing to test.")
+      info("If this is not intended, please check your configuration and try again.")
     }
   }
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala/stryker4s/mutants/Mutator.scala
```diff
@@ -44,12 +44,17 @@ class Mutator(mutantFinder: MutantFinder, transformer: StatementTransformer, mat
...
 
     info(s"Found ${mutatedFiles.size} of $totalAmountOfFiles file(s) to be mutated.")
-    info(s"${includedMutants + excludedMutants} Mutant(s) generated.")
+    info(s"$totalMutants Mutant(s) generated.")
     if (excludedMutants > 0) {
       info(s"Of which $excludedMutants Mutant(s) are excluded.")
     }
+    if (includedMutants == 0) {
+      info("It's a mutant-free world, nothing to test.")
+      info("If this is not intended, please check your configuration and try again.")
```
legopiraat: Maybe we should add some hints on what could be wrong for example:
* Amount of generated files = 0 so file collection might be wrong or the basepath
* Files are found but nothing is mutated maybe everything is excluded

etc.

hugo-vrijswijk: Good point 👍. I'll see if I can rewrite it a little bit tomorrow

hugo-vrijswijk: I've changed the logs a bit. It now logs more about scenarios that could be problematic and provides a hint as to what configuration could be the cause.

PR #142, opened 2019-01-29
Pattern match mutator added.

### LINKED ISSUE #61: Pattern match mutator
### Summary

Am not 100% sure if this is currently covered but I don't think it is, but we could introduce a `Pattern Match` mutator (thinking of a switch statement mutator in our languages).

For example;

```Scala
trait FooBar
case class Foo(name: String = "Foo") extends FooBar
case class Bar(name: String = "Bar") extends FooBar
case class Otherwise() extends FooBar

def fooFunction(foobar: FooBar): Unit = {
  foobar match {
    case foo: Foo => println(foo.name)
    case bar: Bar  => println(bar.name)
    case _             => println("Unknown")
  }
}

fooFunction(Bar())
```

We could remove specific `cases` and have that as a mutator.
The only one potential issue with `sealed traits`.

```Scala
trait FooBar
case class Foo(name: String = "Foo") extends FooBar
case class Bar(name: String = "Bar") extends FooBar
case class Otherwise() extends FooBar

def fooFunction(foobar: FooBar): Unit = {
  foobar match {
    case foo: Foo => println(foo.name)
    case bar: Bar  => println(bar.name)
  }
}

fooFunction(Bar())
```

This example would give a compiler `warning`:`match may not be exhaustive. It would fail on the following input: Otherwise()` by default and the code won't break. But if you add the `-Xfatal-warnings` compiler option it will. This will be a problem for Stryker4s currently.

Wmaarts:
#### What it does
Mutates `match` expressions in the following way: 
For every case clause a new mutant is generated where that clause is omitted from the pattern match statement.

#### How it works
Because removing case clauses can cause compiler warnings which can be flagged as errors, we add the `@unchecked` annotation to the mutated pattern match. ([more info on this annotation here](https://www.scala-lang.org/api/2.12.1/scala/unchecked.html))

So:
```
x match {
  case Some(_) => ???
  case None => ???
}
```
Generates the following two mutations: 
```
(x: @unchecked) match {
  case None => ???
}
```
and
```
(x: @unchecked) match {
  case Some(_) => ???
}
```

Also I reworked the structure of the `SubstitutionMutation` trait slightly, since the new `PatternMatchExpression` trait has unique functionality.

#### Possible (?) additions
- Change the order of case statements 
   - Might be redundant, and could be really expensive if we want to mutate all possibilities.
- Editing the guards on case statements
   - Although this could result in duplicate case clauses


core/src/main/scala/stryker4s/extension/mutationtype/MutationTypes.scala
```diff
@@ -18,7 +18,8 @@ object Mutation {
     classOf[BooleanLiteral].getSimpleName,
     classOf[LogicalOperator].getSimpleName,
     classOf[StringLiteral[_]].getSimpleName,
-    classOf[MethodExpression].getSimpleName
+    classOf[MethodExpression].getSimpleName,
+    classOf[PatternMatchExpression].getSimpleName
```
hugo-vrijswijk: I don't think we should add an extra category. That means we won't follow the [category standard](https://github.com/stryker-mutator/stryker-handbook/blob/master/mutator-types.md#supported-mutators) anymore. I am not sure which category it should be in instead. Perhaps block expression

core/src/test/scala/stryker4s/mutants/findmutants/MutantMatcherTest.scala
```diff
@@ -441,5 +441,33 @@ class MutantMatcherTest extends Stryker4sSuite with TreeEquality {
...
+    }
+
+    it("should not mutate a match with only one case") {
+      val tree = q"object c { variable match { case 1 => 1 } }"
+
+      val found: Seq[Option[Mutant]] = tree.collect(sut.allMatchers).flatten
+
+      found should have length 0
+    }
+
```
hugo-vrijswijk: Could you also add a test case for a PartialFunction? In my experience there can be some subtle differences with the two sometimes

hugo-vrijswijk (changes requested):

core/src/main/scala/stryker4s/extension/mutationtype/PatternMatch.scala
```diff
@@ -0,0 +1,24 @@
...
+    if (tree.cases.length > 1) {
+      Some(
+        tree,
+        tree.cases.map { caze =>
+          import scala.meta.quasiquotes.XtensionQuasiquoteTerm
+
+          // https://www.scala-lang.org/api/2.12.1/scala/unchecked.html
+          // @unchecked is required because we are removing cases here,
+          // possibly resulting in compile warnings (which could be flagged as errors)
+          val annotationWrap = q"${tree.expr}: @unchecked"
```
legopiraat: I really don't like that we are just flagging this as unchecked. 

When sealed traits, for example, the compiler error is already a guard against that mutator. Cases can't be accidentally removed anymore in that case. We could argue the statement in the pattern match (the executing code) is then also removed, that is very true and that should be tested. But that is move removing the actual block statement then removing the case as we do with here.

So in a case of a sealed trait and the code doesn't compile I think Stryker4s should be able to handle that and see that as a killed mutator since Scala is so strongly typed. I know this deviates from the normal stryker implementation but i could see that being different for Stryker4s what do you think @hugo-vrijswijk  @nicojs.

legopiraat (changes requested):

core/src/test/scala/stryker4s/mutants/findmutants/MutantMatcherTest.scala
legopiraat: Plus i would like a more real life scenario test generally you will not be pattern matching numbers, but more likely types.

legopiraat:
I would rather not merge this until we know the impact and more testing scenarios during a full integration test. For me, there are a lot of unknown were potential errors might pop up for example compiler settings etc.

Wmaarts:
> I would rather not merge this until we know the impact and more testing scenarios during a full integration test. For me, there are a lot of unknown were potential errors might pop up for example compiler settings etc.

I agree. We should be careful about the impact of this, it's tricky.

But I should say running stryker4s over itself with this does already work and I did found a couple of things while experimenting. (#143 for example :') )

core/src/main/scala/stryker4s/extension/mutationtype/MutationTypes.scala
Wmaarts: I renamed it to `RemoveStatement` for now. Should probably discuss further what we want to call this.

hugo-vrijswijk:

PR #145, opened 2019-01-30
Add inSubDir extension to get new tmpDir paths

hugo-vrijswijk:
#### What it does

This PR adds an extension method to `File` to get a directory relative to a parameter, instead of relative to the base directory of a project. For example: `File("projectRoot/src/main").inSubDir(File("projectRoot/target/tmp")` returns `projectRoot/target/tmp/src/main`. This is useful for the sbt plugin, which can use it to set the directory for mutated source files. It will also be useful for the maven plugin, which needs do to the same (I made this while working on that, but figured it could be in it's own PR).

#### How it works

It gets the relative path of a File (`src/main`), then prepends the given parameter to it (`projectRoot/target/tmp`).


runners/sbt/src/main/scala/stryker4s/sbt/SbtMutantRunner.scala
```diff
@@ -33,8 +34,7 @@ class SbtMutantRunner(state: State, processRunner: ProcessRunner, sourceCollecto
 
   private val settings: Seq[Def.Setting[_]] = Seq(
     fork in Test := true,
-    scalaSource in Compile := tmpDirFor(Compile).value,
-    scalaSource in Test := tmpDirFor(Test).value,
+    scalaSource in Compile := tmpDirFor(Compile).value
   ) ++
```
Wmaarts: `scalaSource in Test` is no longer necessary?

hugo-vrijswijk: We don't mutate test sources, so there's no point in setting them to the new tmpDir

Wmaarts: Right, good catch 😅

PR #146, opened 2019-01-30
Test formatting in Travis

hugo-vrijswijk:
I've added a bootstrap script for Scalafmt. This allows us to run/check formatting easily. I've also added a check to the travis build.

Why this and not just the sbt plugin? Well, the sbt plugin only covers files in the sbt project (duh). If we want to format other files (like maven projects, sbt scripted projects, etc) this is what we need.


bin/scalafmt.ps1
```diff
@@ -0,0 +1 @@
+&java -jar "$pwd\bin\scalafmt" $args
```
legopiraat: Since we are running stryker4s on a linux server it would make more to make a bash script

legopiraat:
So if i understand correctly it will break the build when the formatting isn't correct. Maybe we also need some webhook to comment on the PR that formatting is wrong or something?

bin/scalafmt.ps1
hugo-vrijswijk: The file it calls is also a Linux shell script. This is just for Windows compatibility

hugo-vrijswijk:
Hmm, maybe. Not sure how to set that up though. Travis does show that the "Formatting" stage failed, and logs which files

legopiraat:
Let's see if this enough if it get's tedious we could also set it as an optional step i assume.

hugo-vrijswijk:
Wild idea: would be pretty cool to have CI autoformat and commit that. Though CI creating commits can be a dangerous thing...

legopiraat:
Well that would be "fancy" but then I would then rather have it open a pull request with the changes like `dependabot` or something.

PR #150, opened 2019-02-07
Logging improvements

### LINKED ISSUE #143: Testing debug logs doesn't work with 'sbt test'
### Summary
I wrote and ran a test with `shouldBe loggedAsDebug`. It passes when running it in IntelliJ.

But this test fails when running `sbt test` from command line. (and so it also fails when running `sbt stryker`)

### How it was found
I figured out that the `case Failure(exception) => ` body in `MatchBuilder.buildnewSource` was not tested properly while experimenting with pattern match mutations.

So I added this test `MatchBuilderTest.scala`, which led me to finding this bug:
```
    it("should log failures correctly") {
      // Arrange
      implicit val ids: Iterator[Int] = Iterator.from(0)
      val source = """class Foo { def foo = true }""".parse[Source].get

      val firstTransformed = toTransformed(source, EmptyString, Lit.Boolean(true), Lit.Boolean(false))

      val transformedStatements = SourceTransformations(source, List(firstTransformed))
      val sut = new MatchBuilder {
        override def buildMatch(transformedMutant: TransformedMutants): Term.Match =
          throw new Exception()
      }

      // Act
      val expectedException = the[UnableToBuildPatternMatchException] thrownBy sut.buildNewSource(transformedStatements)

      // Assert
      "Failed to construct pattern match: original statement [true]" shouldBe loggedAsError
      "Failed mutation(s) Mutant(0,true,false,EmptyString)." shouldBe loggedAsError
      "at Input.String(\"class Foo { def foo = true }\"):1:23" shouldBe loggedAsError
      "This is likely an issue on Stryker4s's end, please enable debug logging and restart Stryker4s." shouldBe loggedAsError

      // The test fails on the line below
      "Please open an issue on github: https://github.com/stryker-mutator/stryker4s/issues/new" shouldBe loggedAsDebug
      "Please be so kind to copy the stacktrace into the issue" shouldBe loggedAsDebug
    }
```

Wmaarts:
##### In the core project 
- Added a unit test in `MatchBuilderTest.scala` as mentioned in #143.
- Removes `log-level` setting from config in favor of 'inheriting' these from the environment. (sbt or maven)
  - This configuration setting was what caused #143.
- Disable log4j logging to console for tests. (now only uses the TestAppender)

##### In the sbt plugin
- Introduces a new way of setting the loglevel in sbt => `logLevel in stryker := Level.Debug`.
- By default disables logging of test results inside the initial test run and inside mutations. (enabled by setting loglevel in stryker to debug)

#### Why do we need `logLevel in Stryker`?
Because unlike maven, sbt doesn't change the log level for the stryker4s plugin if you run it with for example `-debug`. 🙃

#### How it works
It's complicated. 🤔

There doesn't seem to be a clean way of implementing our log4j architecture in the sbt plugin. Sbt uses its own logging framework. Therefore in order to get stryker4s logging we need to activate log4j. We do this by calling `Configurator.setRootLevel` like we used to do in `ConfigReader`.
Right now this leaves us with two seperate log levels: sbt itself and stryker4s. (which is possibly a good thing, have you tried running sbt with the `-debug` flag? my console gets flooded).

Also, the only way I could find to disable logging inside the initial test run and mutation runs was simply overriding the output to an empty `LogManager` (see `emptyLogManager` variable). sbt has no `Level.OFF` like log4j has.

#### State of the PR
~This is still experimental. I wanted to share what I had so far to possibly get some tips and discuss the options.~


legopiraat:
Don't we still need log level for the process runner?

Wmaarts:
> Don't we still need log level for the process runner?

Right now that's correct. But we can also pass flags to change the log level, similar to how it'll work in Maven and sbt.

runners/sbt/src/main/scala/stryker4s/sbt/Stryker4sPlugin.scala
```diff
@@ -15,17 +16,29 @@ object Stryker4sPlugin extends AutoPlugin {
   override def trigger = allRequirements
 
   object autoImport {
+    val Stryker = config("stryker") extend Test
```
hugo-vrijswijk: Are you sure we need a new Configuration for this? `logLevel in stryker` (the task) also compiles

Wmaarts: You are very correct, we don't need this.

core/src/test/resources/log4j2-test.xml
```diff
@@ -2,14 +2,14 @@
 <Configuration packages="stryker4s" status="WARN">
     <Appenders>
         <TestAppender name="TestAppender"/>
-        <Console name="Console" target="SYSTEM_OUT">
-            <PatternLayout pattern="%d{HH:mm:ss.SSS} %-5level %logger{0}: %msg%n"/>
-        </Console>
+        <!--<Console name="Console" target="SYSTEM_OUT">-->
+            <!--<PatternLayout pattern="%d{HH:mm:ss.SSS} %-5level %logger{0}: %msg%n"/>-->
+        <!--</Console>-->
```
hugo-vrijswijk: Why is the console appender removed? Seeing log output during tests can still be useful

runners/command-runner/src/main/scala/stryker4s/run/Stryker4sArgumentHandler.scala
```diff
@@ -0,0 +1,23 @@
...
+import org.apache.logging.log4j.core.config.Configurator
+
+object Stryker4sArgumentHandler extends Logging {
+  private lazy val logLevels: Map[String, Level] =
+    Level.values().map(level => (level.toString.toLowerCase, level)).toMap
+
+  def parseArgs(args: Array[String]): Unit = {
+    // Collect and handle log level argument
+    args
+      .filter(_.head == '-')
```
hugo-vrijswijk:
```suggestion
      .filter(_.startsWith("-"))
```

runners/command-runner/src/main/scala/stryker4s/run/Stryker4sArgumentHandler.scala
```diff
@@ -0,0 +1,23 @@
...
+      .map(_.drop(1))
+      .collectFirst { case arg if logLevels.contains(arg) => setLogLevel(logLevels(arg)) }
+      .getOrElse(setLogLevel(Level.INFO))
```
hugo-vrijswijk: To me this whole function is a bit difficult to understand what's going on. The `collectFirst` would make more sense as a `filter` and a `getOrElse` or something. Also, it's more common to start arguments with two dashes (so, --debug). I think it'd also be nice to have the comparison case-insensitive. I'm also missing tests.

runners/sbt/src/main/scala/stryker4s/sbt/Stryker4sPlugin.scala
```diff
@@ -20,17 +22,38 @@ object Stryker4sPlugin extends AutoPlugin {
...
+
     val result = new Stryker4sSbtRunner(currentState).run()
 
     result match {
       case ErrorStatus => throw new MessageOnlyException("Mutation score is below configured threshold")
       case _           => currentState
     }
   }
+
+  private def sbtLogLevelToLog4j(level: Level.Value): org.apache.logging.log4j.Level = {
```
hugo-vrijswijk: Minor nitpick, but IMO it's nicer to remove the braces on a function if it is only a pattern match

Wmaarts:
So a little update on the state of this PR. 

Currently to set loglevel to DEBUG (or any level) in:
- sbt plugin => use `logLevel in stryker := Level.DEBUG`
- command runner => pass `-debug` as parameter
- ( maven plugin => `-X` should work @hugo-vrijswijk? )

Also a lot of logging is removed by default from the sbt plugin. See [this gist](https://gist.github.com/Wmaarts/b216f7a16ad190564c5c76886a1af5ad) for an example of logging output right now. This is kind of what I hoped to achieve with this PR: no more log cluttering from sbt. 😁

Two things:
- This definitely needs a review. 👀 
- We should still discuss if this is how we want the configuration.

core/src/test/resources/log4j2-test.xml
Wmaarts: The issue I have here: it's useful to show log output when just running tests, but when running stryker4s on itself it results in a lot of (unnecessary?) logging because it logs for every mutation. :/

That's why I disabled it, but I guess I shouldn't have committed it.

runners/command-runner/src/main/scala/stryker4s/run/Stryker4sArgumentHandler.scala
Wmaarts: I refactored it to use `find` instead of `collectFirst`. And also the two dashes and the tests. :)

core/src/test/resources/log4j2-test.xml
legopiraat: I would prefer that we would have logging during tests. when running stryker on stryker we could of course overwrite the loglevel ;)

runners/command-runner/src/main/scala/stryker4s/run/Stryker4sArgumentHandler.scala
```diff
@@ -0,0 +1,26 @@
+package stryker4s.run
+import grizzled.slf4j.Logging
+import org.apache.logging.log4j.Level
+import org.apache.logging.log4j.core.config.Configurator
+
+object Stryker4sArgumentHandler extends Logging {
+  private lazy val logLevels: Map[String, Level] =
+    Level.values().map(level => (level.toString.toLowerCase, level)).toMap
+
+  def parseArgs(args: Array[String]): Unit = {
```
legopiraat: Can we find a better name for this? it's not only parsing args its also setting the loglevel

runners/sbt/src/main/scala/stryker4s/sbt/SbtMutantRunner.scala
```diff
@@ -31,10 +33,16 @@ class SbtMutantRunner(state: State, processRunner: ProcessRunner, sourceCollecto
     }
   }
 
+  private lazy val emptyLogManager = LogManager.defaultManager(ConsoleOut.printStreamOut(new PrintStream((_: Int) => {})))
```
legopiraat: What do you mean with empty log manager?

legopiraat:
Keep in mind that this has an impact on the quickstart on the website in section 4 of the quickstart text.

http://stryker-mutator.io/stryker4s/quickstart

runners/command-runner/src/main/scala/stryker4s/run/Stryker4sArgumentHandler.scala
```diff
@@ -0,0 +1,26 @@
+package stryker4s.run
+import grizzled.slf4j.Logging
+import org.apache.logging.log4j.Level
+import org.apache.logging.log4j.core.config.Configurator
+
+object Stryker4sArgumentHandler extends Logging {
+  private lazy val logLevels: Map[String, Level] =
```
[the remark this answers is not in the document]
Wmaarts: If no arguments are passed to the `CommandRunner`, this is never called, so not necessary to fill the `val`.

runners/sbt/src/main/scala/stryker4s/sbt/SbtMutantRunner.scala
Wmaarts: Basically it swallows logging. It doesn't do anything with the logging it receives. Does it need a different name?

runners/command-runner/src/main/scala/stryker4s/run/Stryker4sArgumentHandler.scala
Wmaarts: I changed it to `handleArgs` for now. (which should make sense because the class is called a `Handler` too 😅 )

core/src/test/resources/log4j2-test.xml
cbrunnkvist: It is difficult to understand what exactly the tool is doing / which tests are actually being executed, and if there is no way to see the regular test report, then how can I gain any confidence in the mutation reports? #DevEx

PR #151, opened 2019-02-07
Maven plugin

### LINKED ISSUE #141: Maven plugin
I've started work on a Maven plugin in the [maven-plugin](https://github.com/stryker-mutator/stryker4s/tree/maven-plugin) branch. Although sbt still accounts for [most Scala projects](https://www.jetbrains.com/research/devecosystem-2018/scala/), Maven is not far off. Most projects (non open source) that i have seen within work also use Maven.

With that in mind, a couple things should be noted about a Maven plugin:

- Naming. If the plugin is in the format `foo-maven-plugin` and a task (mojo) named `foo` is available, it allows the user to just execute `mvn foo`. Otherwise, more information needs to be included. Right now the plugin is named `stryker4s-maven-plugin` with the mojo named `stryker`. This means the plugin can be called with `mvn stryker4s:stryker`. IMO this is not ideal. We could change the mojo name to `stryker4s`, which would be out of line with the sbt plugin (which is `sbt stryker`), or rename the plugin to `stryker-maven-plugin`, which could be confusing as it makes it look like the plugin is for Stryker JS. EDIT: Having the mojo as one statement (`mvn stryker`) is not possible, as far as I can tell. So the command would be either [`mvn stryker4s:stryker4s`](https://en.wiktionary.org/wiki/Department_of_Redundancy_Department) or `mvn stryker4s:stryker`
- Project. It's not possible (as far as I can tell) to build a maven plugin in sbt. This means the maven plugin is a separate project. If we keep the project in the same repo (which I think makes the most sense), this means extra build configuration for releases (and snapshots?). 
- Scala versions. The maven plugin is built with Scala 2.12.8. I don't know if this has impact on projects that run on 2.11. Something we need to test.

hugo-vrijswijk:
#### What it does

This PR adds a Maven plugin, similar to the sbt plugin. The project is named `stryker4s-maven-plugin` and exposes a single Mojo (task) named `run`, which can be invoked with `stryker4s:run`.

As it is not possible (unless someone can surprise me here) to build a maven plugin from sbt, it is added as a separate Maven project in `runners/maven`. Unfortunately, this means some stuff (like dependency versions, dev info, SCM info, etc) will have to be duplicate. 

It also means extra work for releases. Publishing isn't in this PR (yet), but I wanted to put the rest up for review/discussion first. 

#### How it works

The [maven-invoker](https://maven.apache.org/shared/maven-invoker/) plugin is used to invoke the `test` goal for each mutant. The base-directory for the invoker is set to the temp directory. This means a new process is started. Unfortunately, we don't have any control over the process (similar to sbt plugin).

I've tested the plugin with a small project (1 class, 1 test), but would like to test it on some bigger projects before releasing.

#### Known issues

We meet again, JVM properties. Properties are not copied over to mutation runs (`mvn stryker4s:run -DargLine="-Dconfig.file=foo"`).


hugo-vrijswijk:
I have added the build steps. It was a bit more work than I would've liked, but it works nicely now IMO. Maven releases are only called when sbt does a release, and uses the same version.

It customizes the publish command that `sbt-ci-release` calls by changing an environment variable (see: `travis.yml`). Some new commands are added to the build settings and defined in `project/Release.scala`. The `stryker4sPublish` and `stryker4sPublishSigned` commands are used by `sbt-ci-release` instead of the normal `+publishSigned` (releases) or `+publish` (snapshots). 

The flow is as follows:
1. `stryker4sMvnSetup`: 
    1. `publishM2`: Do an install of stryker4s-core in the local maven repository so that the maven plugin can find it as a dependency
    2. `stryker4sMvnPackage`: Set the version of the maven project to the version from `sbt-dynver`
    3. `stryker4sMvnPackage`: Package the project, ensuring that it compiles and builds
    4. `setVersion`: Set the version in sbt again to work around a problem [we created ourselves](https://github.com/stryker-mutator/stryker4s/blob/69e6a62a059e845318cbb2e3c174aa4bc2aaaf97/project/Release.scala#L64-L67)
2. `+publish(Signed)`: Publish sbt projects, sign if release (unchanged)
3. `stryker4sMvnDeploy`: Call `mvn deploy` from the maven project.

If one step fails, the next ones aren't executed (using the `State` object from sbt).

As for passwords, secrets, etc: The same environment variables are used. Sonatype secrets are imported from environment variables by using a local `settings.xml`. The settings file location is specified from the deploy command. PGP secrets to sign the build are defined in the `pom.xml` maven-pgp-plugin configuration (and also retrieved from the same environment variables as `sbt-ci-release`).

project/Release.scala
```diff
@@ -0,0 +1,69 @@
...
+    *
+    * @param command: The command to run in the maven project
+    */
+  private def mvnGoal(command: String, baseDir: File): process.ProcessBuilder =
+    process.Process(s"mvn --batch-mode $command -P release", baseDir / "runners" / "maven")
+
+  /** After setting the version in the sbt project, there will be local git changes.
+    * this causes `sbt-dynver` to add the timestamp to the end of the version.
+    * To get around this, the version is remembered from the start, and set again after any maven goals
+    */
```
Wmaarts: Feels like this should be a comment instead of a scaladoc.

project/Release.scala
```diff
@@ -0,0 +1,69 @@
...
+    * then set new version and run `mvn install`
+    */
+  private def mvnPackage(version: String, baseDir: File): State => State = handleProcessResult(
+    _,
+    mvnGoal(s"versions:set -DnewVersion=$version", baseDir) #&&
+      mvnGoal("package", baseDir)
+  )
+
+  /** Deploy the maven plugin project
+    */
```
Wmaarts: Docs like these feel a bit redundant. This is basically a more verbose version of the function name.

I don't really understand why there's so many scaladoc here. 🤔

runners/maven/src/main/scala/stryker4s/run/MavenMutantRunner.scala
```diff
@@ -0,0 +1,60 @@
...
+    val request = createRequest
+
+    val result = invoker.execute(request)
+
+    result.getExitCode == 0
+  }
+
+  override def runMutant(mutant: Mutant, workingDir: File, subPath: Path): MutantRunResult = {
+    val request = createRequestWithMutation(mutant)
+    request.addShellEnvironment("ACTIVE_MUTATION", String.valueOf(mutant.id))
```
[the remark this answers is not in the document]
hugo-vrijswijk: Correct!

project/Release.scala
hugo-vrijswijk: Hmm. Yeah probably. I've changed it for now

project/Release.scala
hugo-vrijswijk: Yeah I went a bit scaladoc-crazy. It looked more complex at first.

runners/maven/src/main/scala/stryker4s/run/Stryker4sMavenPlugin.scala
```diff
@@ -0,0 +1,22 @@
...
+import org.apache.maven.plugins.annotations.{Mojo, Parameter}
+import org.apache.maven.project.MavenProject
+import stryker4s.run.threshold.ErrorStatus
+
+/** The main goal for this plugin. Starts Stryker4s.
+  */
+@Mojo(name = "run")
+class Stryker4sMavenPlugin @Inject()(@Parameter(defaultValue = "${project}") project: MavenProject)
+    extends AbstractMojo
+    with Logging {
```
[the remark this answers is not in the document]
hugo-vrijswijk: Nope

runners/maven/src/main/scala/stryker4s/run/MavenMutantRunner.scala
```diff
@@ -0,0 +1,56 @@
...
+
+    val result = invoker.execute(request)
+
+    result.getExitCode match {
+      case 0 => Survived(mutant, _)
+      case _ => Killed(mutant, _)
+    }
+  }
+
+  private def createRequest: InvocationRequest =
```
legopiraat: createRequest() please to make it clear is a func

runners/maven/src/main/scala/stryker4s/run/MavenMutantRunner.scala
```diff
@@ -0,0 +1,56 @@
...
+) extends MutantRunner(sourceCollector) {
+
+  private val goals = List("test").asJava
+
+  private val properties = new Properties(project.getProperties)
+  properties.setProperty("surefire.skipAfterFailureCount", 1.toString) // Stop after first failure. Only works with surefire plugin, not scalatest
+
+  override def runInitialTest(workingDir: File): Boolean = {
+    // Set source once, settings is persistent among goals
+    invoker.setWorkingDirectory(workingDir.toJava)
```
legopiraat: putt this on class level? doesn't make sense to do it initialtestRun if it's persistent. If we would ever disable the initial test run this would break the mutation run

project/Release.scala
```diff
@@ -0,0 +1,59 @@
...
+      Command.command(stryker4sPublish)(stryker4sMvnSetup :: `+publish` :: stryker4sMvnDeploy :: _),
+      Command.command(stryker4sPublishSigned)(stryker4sMvnSetup :: `+publishSigned` :: stryker4sMvnDeploy :: _),
+      // Called by stryker4sPublish(signed)
+      Command.command(stryker4sMvnSetup)(publishM2 :: stryker4sMvnPackage :: setVersion(versionNumber) :: _),
+      Command.command(stryker4sMvnPackage)(mvnPackage(versionNumber, baseDirectory.value)),
+      Command.command(stryker4sMvnDeploy)(mvnDeploy(baseDirectory.value))
+    )
+  }
+
+  private def mvnPackage(version: String, baseDir: File): State => State = handleProcessResult(
```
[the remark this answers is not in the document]
hugo-vrijswijk: True. I did this as a first step so that you wouldn't run into the scenario of publishing the sbt project, and then finding out the maven project didn't compile or something. I've now added a test stage, so this is a bit redundant. I'll remove this and set the version in mvnDeploy

PR #153, opened 2019-02-12
Eta calculator

Wmaarts:
#### What it does
Logs estimated time remaining after each run. For example:
`INFO  stryker4s.sbt.SbtMutantRunner - Finished mutation run 2/9 (22%). 0 minutes and 55 seconds remaining.`
Optionally includes hours if the estimate is over an hour.

#### How it works
Measures the length of every run and bases the estimation on the last 5 runs. Does not use the initial test run because it seems to take considerably longer.


core/src/main/scala/stryker4s/run/MutantRunner.scala
```diff
@@ -75,18 +75,24 @@ abstract class MutantRunner(process: ProcessRunner, sourceCollector: SourceColle
     filePath.overwrite(mutatedFile.tree.syntax)
   }
 
-  private def runMutants(mutatedFiles: Iterable[MutatedFile], tmpDir: File): Iterable[MutantRunResult] = {
+  private def runMutants(mutatedFiles: Iterable[MutatedFile],
+                         tmpDir: File): Iterable[MutantRunResult] = {
     val totalMutants = mutatedFiles.flatMap(_.mutants).size
+    val etaCalculator = new EtaCalculator(totalMutants)
```
hugo-vrijswijk: It's probably nicer to have this constructor injected, and in the test simply test if the function is called with the correct parameters. Then the EtaCalculator can test if the correct things are logged. IoC and all that

core/src/main/scala/stryker4s/run/EtaCalculator.scala
```diff
@@ -0,0 +1,43 @@
+package stryker4s.run
+
+import grizzled.slf4j.Logging
+
+import scala.concurrent.duration._
+import language.postfixOps
+
+class EtaCalculator(amountOfMutants: Int) extends Logging {
```
hugo-vrijswijk: I like the solution, but I think I would prefer if the EtaCalculator only calculated the ETA. Then the result could be used in different ways. Now it both times and logs

core/src/main/scala/stryker4s/run/EtaCalculator.scala
```diff
@@ -0,0 +1,43 @@
...
+    prettyPrintDuration((amountOfMutants - currentRun) * average milliseconds)
+  }
+
+  private def prettyPrintDuration(duration: Duration): String = {
+    val durationInSeconds = duration.toSeconds
+    val hours = durationInSeconds / 3600
+    val minutes = (durationInSeconds % 3600) / 60
+    val seconds = durationInSeconds % 60
+
+    (if(hours > 0) s"$hours hours " else "" ) + s"$minutes minutes and $seconds seconds"
```
hugo-vrijswijk: How much sense does it make to have an ETA for seconds? Unless we make some big performance improvements. It's up for discussion, but to make having only seconds (and hours) makes more sense, and something like "< 1 minute" for under a minute

hugo-vrijswijk (changes requested):

core/src/main/scala/stryker4s/run/EtaCalculator.scala
```diff
@@ -0,0 +1,43 @@
...
+
+    // Log estimate
+    log(calculateETA(runNr + 1))
+
+    result
+  }
+
+  def calculateETA(currentRun: Int): String = {
+    val from = math.max(0, currentRun - 5) // Base estimate on last 5 runs
+    val average = runResults.slice(from, currentRun).sum / (currentRun - from)
```
hugo-vrijswijk: Why the average from the last 5 runs? If a couple of the last runs time out this will really skew the results. The median of all runs makes more sense to me.

Wmaarts: The idea with taking the last 5 runs is basically we say: "If things keep going like this, it will take X minutes". Taking the median of all runs could result in skewed results as well. 

For example: What if halfway through a run you boot up a different program that eats up most of your RAM and CPU? Stryker will slow down significantly, but the estimate will take a long time to adjust because you've already done so many runs, so the average will hardly change. Resulting in a estimate consistently lower than the actual time. 

Maybe 5 runs is not enough, maybe it should be 10? Still I would argue this is better than taking the median.

We would have to think of (or find) some formula that takes all runs in consideration but does not count the first few runs with equal 'severity' as the more recent runs if we'd really want to do this properly. 🤔

core/src/main/scala/stryker4s/run/MutantRunner.scala
Wmaarts: I agree, but I'm not sure how I should handle passing `totalMutants` then, since it's not available at time of constructing. =/

Wmaarts: Perhaps a `start(totalMutants)` function that returns a new EtaCalculator or something...?

core/src/main/scala/stryker4s/run/EtaCalculator.scala
Wmaarts: I changed it slightly to just return the result of the function and the estimate. Maybe that's better?

core/src/main/scala/stryker4s/run/EtaCalculator.scala
hugo-vrijswijk: Hmm, good point. I do feel like 5 is too little. 10 is probably good. I do still feel like the median of the last 10 would be better than the average. To filter out any timeouts/outliers.

core/src/main/scala/stryker4s/run/EtaCalculator.scala
```diff
@@ -0,0 +1,40 @@
...
+    val duration = System.currentTimeMillis() - startTime
+    val runNr = stream.next()
+    runResults(runNr) = duration
+
+    (result, calculateETA(runNr + 1))
+  }
+
+  def calculateETA(currentRun: Int): String = {
+    val from = math.max(0, currentRun - 5) // Base estimate on last 5 runs
+    val average = runResults.slice(from, currentRun).sum / (currentRun - from)
```
hugo-vrijswijk: BTW you can also use `takeRight(n)` to to take the last n elements, or less if the array is smaller.

Wmaarts: I would use that if this was a Stack that grows in size, but it's an Array. 😁

hugo-vrijswijk: Ah duh. So why not use a mutable.List?

Wmaarts: Because we know beforehand how long the List is going to be. I believe in that case Array is slightly faster.

hugo-vrijswijk: Hmm, true. Thinking about it. Wouldn't a queue (or whatever) with a max size of 10 (median/average size) be more efficient? Pushing out the oldest element as a new one is added.

core/src/main/scala/stryker4s/run/EtaCalculator.scala
Wmaarts: Personally I don't think it matters all that much. Both options are fine. :)

core/src/main/scala/stryker4s/run/EtaCalculator.scala
Wmaarts: I changed it to 10 and used median instead of average. 👍

core/src/main/scala/stryker4s/run/EtaCalculator.scala
```diff
@@ -0,0 +1,54 @@
+package stryker4s.run
+
+import grizzled.slf4j.Logging
+
+import scala.concurrent.duration._
+import language.postfixOps
+import scala.collection.mutable
+
+class EtaCalculator(amountOfMutants: Int) extends Logging {
```
legopiraat: In general, I get the feeling that this class is doing a lot of stuff it's not responsible for.

1. prettyPrintDuration could be an extension method on Duration so I expected an implicit class for the duration in that case.
2. No way the ETACalculator is responsible for invoking the "fun" because that seems to be the "runMutant" function
3. keeping track of the duration a mutation run took is no responsibility for this class. the runMutant function could result in mutationResult which includes a duration.

When considering option 3 this class could then simply queue the last 10 and calculate an avg based on the runtime. This avg is then reported as a duration, which can then be used in the mutantrunner to be prettyPrinted using the extentionMethod.

legopiraat (changes requested):

core/src/main/scala/stryker4s/run/EtaCalculator.scala
Wmaarts: To be fair, I had a feeling I wasn't thinking according to the correct paradigm here either...

So please help me understand a few things. :')

1. I can get behind this one... I suppose conversion/parsing logic like this is cleaner in a seperate class. The reason I didn't do it initially was because: "it's only used here, just make it a private function".
1. It is basically calling the `runMutant` function, true. But I felt like the alternative is doing something awful like:
```
EtaCalculator.start()
runMutant(...)
EtaCalculator.stop()
```
How do I implement this neatly then? :/ 

3.  Giving `runMutant` responsibility to time itself means we need to implement the same timing logic in every implementation of `MutantRunner` which would result in duplicate code. (Or we would have to introduce another function that calls `runMutant` and wraps it inside a `MutantRunResult` with the duration?)

hugo-vrijswijk:

PR #155, opened 2019-02-13
Add takeRight, dropRight, takeWhile, dropWhile mutators

hugo-vrijswijk:
#### What it does

I've added mutators for `takeRight` -> `dropRight` (and other way around), and `takeWhile` -> `dropWhile` (and other way around).

#### How it works

Mutator magic 🔮


legopiraat:
Don't forget to add them here 👍 
https://github.com/stryker-mutator/stryker-handbook/blob/master/mutator-types.md#stryker4s

hugo-vrijswijk:
@legopiraat Done! See [here](https://github.com/stryker-mutator/stryker-handbook/pull/15)

PR #157, opened 2019-02-14
If statement mutator

Wmaarts:
#### What it does
Mutates if-statements like so:
`if(...)` => `if(false)` and `if(true)`

#### How it works
Matches on a `Term` that is a _direct_ condition of an if-statement and generates two mutations for it: `true` and `false`. 

#### Caveats
With the current implementation it's not (yet) possible to exclude these mutations seperately, since it uses the `BooleanLiteral` mutation.

### DIFF
--- a/core/src/main/scala/stryker4s/extension/TreeExtensions.scala
+++ b/core/src/main/scala/stryker4s/extension/TreeExtensions.scala
@@ -98,15 +98,4 @@ object TreeExtensions {
     final def isIn[T <: Tree](implicit classTag: ClassTag[T]): Boolean =
       mapParent[T, Boolean](thisTree, _ => true, false)
   }
-
-  implicit class IsConditionOfIf(thisTree: Tree) {
-    /** Returns if a tree is a direct condition of an if statement
-      */
-    final def isConditionOfIf: Boolean = {
-      thisTree.parent match {
-        case Some(parent: Term.If) => parent.cond == thisTree
-        case _ => false
-      }
-    }
-  }
 }
--- a/core/src/main/scala/stryker4s/extension/TreeExtensions.scala
+++ b/core/src/main/scala/stryker4s/extension/TreeExtensions.scala
@@ -98,4 +98,15 @@ object TreeExtensions {
     final def isIn[T <: Tree](implicit classTag: ClassTag[T]): Boolean =
       mapParent[T, Boolean](thisTree, _ => true, false)
   }
+
+  implicit class IsConditionOfIf(thisTree: Tree) {
+    /** Returns if a tree is a direct condition of an if statement
+      */
+    final def isConditionOfIf: Boolean = {
+      thisTree.parent match {
+        case Some(parent: Term.If) => parent.cond == thisTree
+        case _ => false
+      }
+    }
+  }
 }
--- a/core/src/main/scala/stryker4s/extension/mutationtype/ConditionalExpression.scala
+++ b/core/src/main/scala/stryker4s/extension/mutationtype/ConditionalExpression.scala
@@ -1,0 +1,16 @@
+package stryker4s.extension.mutationtype
+
+import scala.meta.{Lit, Term}
+
+case object IfStatement {
+  def unapply(tree: Term.If): Option[(Term, Seq[IfStatementMutation])] = {
+    Some(
+      tree,
+      Seq(
+        IfStatementMutation(tree.copy(cond = Lit.Boolean(true))),
+        IfStatementMutation(tree.copy(cond = Lit.Boolean(false))),
+      )
+    )
+  }
+  case class IfStatementMutation(tree: Term.If) extends ConditionalExpression[Term.If]
+}
--- a/core/src/main/scala/stryker4s/extension/mutationtype/ConditionalExpression.scala
+++ b/core/src/main/scala/stryker4s/extension/mutationtype/ConditionalExpression.scala
@@ -9,7 +9,7 @@ case object While {
   def unapply(whileStatement: Term.While): Option[Term] = Some(whileStatement.expr).filterNot(_.is[Lit.Boolean])
 }
 case object Do {
-  def unapply(whileStatement: Term.Do): Option[Term] = Some(whileStatement.expr).filterNot(_.is[Lit.Boolean])
+  def unapply(doStatement: Term.Do): Option[Term] = Some(doStatement.expr).filterNot(_.is[Lit.Boolean])
 }
 
 case object ConditionalTrue extends ConditionalExpression {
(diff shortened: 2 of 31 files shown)


core/src/main/scala/stryker4s/extension/TreeExtensions.scala
```diff
@@ -98,4 +98,15 @@ object TreeExtensions {
     final def isIn[T <: Tree](implicit classTag: ClassTag[T]): Boolean =
       mapParent[T, Boolean](thisTree, _ => true, false)
   }
+
+  implicit class IsConditionOfIf(thisTree: Tree) {
```
hugo-vrijswijk: In my opinion this is fine to keep in the `ConditionalExpression` class. I don't see us using it anywhere else, and there already is a bit of tree logic there.

core/src/main/scala/stryker4s/extension/mutationtype/ConditionalExpression.scala
```diff
@@ -0,0 +1,13 @@
+package stryker4s.extension.mutationtype
+
+import scala.meta.Term
+import stryker4s.extension.TreeExtensions.IsConditionOfIf
+
+case object IfStatementCondition {
+
+  def unapply(term: Term): Option[(Term, Seq[BooleanLiteral])] = {
```
hugo-vrijswijk: Was there a specific reason this is done differently than the other mutators? Where the return is an `Option[Term]` (see boolean mutator). Then the function could be something like:

```scala
term.parent collect {
  case Term.If(cond, _, _) if cond == term => term
}
```

core/src/main/scala/stryker4s/extension/mutationtype/ConditionalExpression.scala
```diff
@@ -0,0 +1,13 @@
...
+    if (term.isConditionOfIf)
+      Some((term, Seq(True, False)))
```
hugo-vrijswijk: We should also filter out if the term is already a `true` or `false, as those are already mutated

core/src/main/scala/stryker4s/mutants/findmutants/MutantMatcher.scala
```diff
@@ -37,6 +38,10 @@ class MutantMatcher()(implicit config: Config) {
     case Or(orig)  => orig ~~> And
   }
 
+  def matchConditionalExpression: PartialFunction[Tree, Seq[Option[Mutant]]] = {
+    case IfStatementCondition(orig, mutatedTerms)  => orig ~~> (mutatedTerms: _*)
```
hugo-vrijswijk: See my above comment, it makes more sense to stay in line where this class is responsible for matching statements to mutators, and the mutationType for finding a node in a tree that can be mutated

core/src/main/scala/stryker4s/extension/mutationtype/ConditionalExpression.scala
Wmaarts: This makes more sense to me too. 😅 👍

Wmaarts:
> What was the reason the boolean literal mutator was used now for the exclusion?

Because the `SubstitutionMutation` used was of type `BooleanLiteral`. But now this is refactored to be of type `ConditionalExpression` so it can be exluded specifically.

core/src/main/scala/stryker4s/extension/mutationtype/ConditionalExpression.scala
```diff
@@ -0,0 +1,18 @@
+package stryker4s.extension.mutationtype
+
+import scala.meta._
+
+case object If {
+
+  def unapply(term: Term): Option[Term] =
+    term.parent collect {
+      case Term.If(cond, _, _) if cond == term => term
```
hugo-vrijswijk:
```suggestion
      case Term.If(`term`, _, _) => term
```

core/src/main/scala/stryker4s/extension/mutationtype/ConditionalExpression.scala
```diff
@@ -0,0 +1,18 @@
+package stryker4s.extension.mutationtype
+
+import scala.meta._
+
+case object If {
+
+  def unapply(term: Term): Option[Term] =
+    term.parent collect {
```
hugo-vrijswijk: Thinking about it a bit more: Why not something like this:

```scala
def unapply(if: Term.If): Option[Term] = Some(if.cond).filterNot(_.is[Lit.Boolean])
```

That way the unapply matches on a Term.If, instead of checking every node if it has an If as a parent. Which is probably not very efficient

Wmaarts: This does result in mutating the entire if-statement instead of just the condition, because then the `orig` in `case If(orig) => orig ~~> (IfTrue, IfFalse)` will be the whole if-statement. We might be able to work around that but I don't immediately see how...

hugo-vrijswijk: Would it? You make an Option of the condition (`Some(ifStatement.cond)`). Wouldn't that result in the `orig` being the condition? And that then being mutated to true & false

Wmaarts: Actually you are correct, just tested it. :') Good catch.

Wmaarts:
> Do you think you could also implement the other [conditional mutators](https://github.com/stryker-mutator/stryker-handbook/blob/master/mutator-types.md#conditional-expression)? Only `for` and `while` I guess, as ternary operator does not exist in Scala. Note that `while` should only be mutated to `false`.

I implemented `while` and `do {} while ()` but the `for` expression is quite a bit different in Scala compared to other languages of Stryker. It should be a seperate PR/issue imo.

core/src/main/scala/stryker4s/extension/mutationtype/ConditionalExpression.scala
```diff
@@ -0,0 +1,20 @@
...
+
+import scala.meta._
+
+case object If {
+  def unapply(ifStatement: Term.If): Option[Term] = Some(ifStatement.cond).filterNot(_.is[Lit.Boolean])
+}
+case object While {
+  def unapply(whileStatement: Term.While): Option[Term] = Some(whileStatement.expr).filterNot(_.is[Lit.Boolean])
+}
+case object Do {
```
hugo-vrijswijk: Could you rename this to DoWhile? Just 'Do' is not very clear

Wmaarts: Fair enough, I only called it Do because that's what the Term is called in scala-meta.

PR #158, opened 2019-02-18
Refactor reporters and log detected mutants

### LINKED ISSUE #156: Log killed mutants in report
It'd be nice to also see the killed mutants, possibly with a config option.

Configuration would be something like this:

```conf
stryker4s {
  reporters: [
    console: {
      log-killed-mutants: true
    }
  ]
}
```

Default would be false? Not sure what is best here

We should also still support the old configuration: `reporters: [ "console" ]`

hugo-vrijswijk:
#### What it does

Functionally: this slightly changes the look of logged mutants. And will log detected mutants on debug. The new logging will look like this:

```
17:04:53.468 [pool-30-thread-1] INFO  stryker4s.run.report.ConsoleReporter - Undetected mutants:
0. [Survived]
src\main\scala\stryker4s\config\ConfigReader.scala:16:66
-       "stryker4s.conf"
+       ""

5. [Survived]
src\main\scala\stryker4s\config\ConfigReader.scala:40:13
-       "Failures in reading config: "
+       ""

6. [Survived]
src\main\scala\stryker4s\config\ConfigReader.scala:52:10
-       s"Set logging level to $level."
+       s""
```

I would've added the mutation category, but I removed from `Mutant` it in #154 🤷‍♂️

#### How it works

I've refactored the logging quite a bit. Reporters are no longer inside the config, instead the config just has simple case classes. The reporters are then mapped to their implementations. Also, `Stryker4s.scala` no longer needs a `Reporter`, instead the `MutantRunner` does.

Reporting is called on mutation start, finish, and when all are finished.


core/src/main/scala/stryker4s/report/MutantRunReporter.scala
```diff
@@ -0,0 +1,12 @@
+package stryker4s.report
+
+import stryker4s.model.{Mutant, MutantRunResult, MutantRunResults}
+
+trait MutantRunReporter {
```
legopiraat: I find this interface very strange, it feels like this was designed with only the console reporter in mind. 
When looking at the HTML reporter this interface doesn't make sense anymore. 

There are two different types of reporters, in my opinion, one for reporting the final result of the complete mutation run. and a second type would be a progress reporter, this progress reporter will report the current status etc. This progress reporter is something we could automatically turn on untill the user maybe turns something like a silent mode on or set logging levels accordingly. 

What are your toughts about this @hugo-vrijswijk ?

hugo-vrijswijk:
Hmm, I agree. My idea was the HTML reporter (or any other) would have an empty implementation for any events that it doesn't want to do anything with. But I guess that is a little weird

core/src/main/scala/stryker4s/report/MutantRunReporter.scala
hugo-vrijswijk: I've refactored the reporter to two separate traits: `ProgressReporter`, and `FinishedRunReporter`. Any reporter can implement the traits and the `CombinedReporter` will call it if it is implemented

PR #163, opened 2019-02-25
Fix topstatement for ! and literals

### LINKED ISSUE #161: Invalid code generated for (!_.foo("bar"))
### Summary

Here is minimisation of the issue:

```scala
object Main {
  def main(args: Array[String]): Unit = {
    List(1, 2, 3).map(!_.toString.startsWith("1"))
  }
}
```

becomes:

```scala
object Main {
  def main(args: Array[String]): Unit = {
    List(1, 2, 3).map(!(sys.props.get("ACTIVE_MUTATION") match {
      case Some("1") =>
        _.toString.startsWith("")
      case _ =>
        _.toString.startsWith("1")
    }))
  }
}
```

```
with two compiler errors:
missing parameter type for expanded function ((x$1: <error>) => x$1.toString.startsWith(""))
[error]         _.toString.startsWith("")
[error]         ^
missing parameter type for expanded function ((x$2: <error>) => x$2.toString.startsWith("1"))
[error]         _.toString.startsWith("1")
[error]         ^
```

This problem probably happens in other cases.
This could be solved by placing the 'ACTIVE_MUTATION' block just around the string like:
```scala
object Main {
  def main(args: Array[String]): Unit = {
    List(1, 2, 3).map(!_.toString.startsWith((sys.props.get("ACTIVE_MUTATION") match {
      case Some("1") =>
        ""
      case _ =>
        "1"
    })))
  }
}
```

### Stryker4s config

default

### Stryker4s environment

```
stryker4s 0.2.1
```

### Your Environment
| software                    | version(s) |
| Scala version            |  2.12.8          |
| Build tool & version  | SBT 1.2.8          | 
| Operating System    | MacOS Mojave 10.14.3           |

hugo-vrijswijk:
#### What it does

Statements with ! are now included in topStatement, and literals are now considered a whole topStatement

#### How it works

Change some of the topStatement logic for literals and if the parent is a `Term.Name`


hugo-vrijswijk:
@stryker-mutator/stryker4s This PR is ready for review

PR #164, opened 2019-02-27
[HTML reporter] Download mutation-test-elements.js as dependency

hugo-vrijswijk:
See https://github.com/stryker-mutator/stryker4s/pull/160#issuecomment-467810217

Adds the mutation-testing-elements as a dependency, instead of adding it in the resources folder and checking into source control.

This uses [webjars](http://webjars.org/), which is a service that mirrors npm dependencies as maven dependencies. Downside to this is for every release of mutation-testing-elements we have to request a new upload on webjars manually. Though perhaps this could be integrated into the build for mutation-testing-elements.

An alternative to this would be to download the report from unpkg as part of the sbt build steps. Though that's likely a bit more work.


legopiraat:
How are `stryker` and `stryker.net` dealing with this?

I feel like this index.html page should never land in a class but more as a resource. But also is it really necessary? cant the web components just deliver a default index.html?

hugo-vrijswijk:
> How are stryker and stryker.net dealing with this?

Stryker can just have it as an NPM dependency. Stryker.NET I am not sure. 

> I feel like this index.html page should never land in a class but more as a resource. But also is it really necessary? cant the web components just deliver a default index.html?

I agree inside a class is a bit weird, but the file is fairly small, so I decided to just put it in there. The reason there's no default index.html is because we set some properties on the element (like the report name), and have to inline the JSON output

nicojs:
>   How are stryker and stryker.net dealing with this?
> Stryker can just have it as an NPM dependency. Stryker.NET I am not sure.

Stryker will just add it as a dependency yes. Stryker.NET also adds it as an npm dependency (with a package.json). They are missing the webjars like functionality.

In the future, I would like mutation-testing-elements to be published to maven, nuget, npm, pub, gems, etc from the mutation-testing-elements build itself. First let met automate the npm release, after that PRs for others are welcome.

> I agree inside a class is a bit weird, but the file is fairly small, so I decided to just put it in there. The reason there's no default index.html is because we set some properties on the element (like the report name), and have to inline the JSON output

That sounds reasonable. Yes inlining JSON output is a necessary evil I'm afraid. However, I do think it looks kind of clean to have just 2 files sitting in the report directory. It's magical 🧙‍♂️

hugo-vrijswijk:
> That sounds reasonable. Yes inlining JSON output is a necessary evil I'm afraid. However, I do think it looks kind of clean to have just 2 files sitting in the report directory. It's magical 🧙‍♂️

You can also inline the js file for even more magic 🧝‍♀️🧙‍♂️🌟🔮

PR #165, opened 2019-03-04
Integration test project

legopiraat:
#### What it does

This is an initial setup for an integration test project. It's a small robobar application based on the stryker robobar. It's writtin an a scala idom way with cats to hopefully use common coding patterns so we can test if we support them.

The project is not done since barely any tests are created and also is missing some more mutation types. But i wanted to get a first version out if i could.

Currently there is a 33% mutation score with 30 mutations generated.


hugo-vrijswijk:
Could you add a separate travis job for the integration tests? Where it installs the current version locally and runs that against the IT project

legopiraat:
@hugo-vrijswijk sounds like a good improvement ill try to add that later.

integration-tests/build.sbt
```diff
@@ -0,0 +1,33 @@
+lazy val root = (project in file("."))
+  .settings(
+    organization := "io.stryker4s",
```
hugo-vrijswijk:
```suggestion
    organization := "io.stryker-mutator",
```

integration-tests/project/Dependencies.scala
```diff
@@ -0,0 +1,27 @@
+import sbt._
+
+object Dependencies {
+
+  object Version {
+    val cats = "1.5.0"
+    val http4s = "0.20.0-M5"
+    val scalatest = "3.0.5"
+    val Logback = "1.2.3"
```
hugo-vrijswijk:
```suggestion
    val logback = "1.2.3"
```

integration-tests/project/Dependencies.scala
```diff
@@ -0,0 +1,27 @@
+import sbt._
+
+object Dependencies {
+
+  object Version {
+    val cats = "1.5.0"
```
hugo-vrijswijk: 1.6.0 is out
```suggestion
    val cats = "1.6.0"
```

integration-tests/project/Dependencies.scala
```diff
@@ -0,0 +1,27 @@
...
+  val circeGeneric = "io.circe" %% "circe-generic" % Version.circeVersion
+  val http4sServer = "org.http4s" %% "http4s-blaze-server" % Version.http4s
+  val http4sCirce = "org.http4s" %% "http4s-circe" % Version.http4s
+  val http4sDsl = "org.http4s" %% "http4s-dsl" % Version.http4s
+  val logback = "ch.qos.logback" % "logback-classic" % Version.Logback
+
+  object Testing {
+    val scalactic = "org.scalactic" %% "scalactic" % Version.scalatest % Test
+    val scalatest = "org.scalatest" %% "scalatest" % Version.scalatest % Test
+    val mockitoScala = "org.mockito" %% "mockito-scala" % Version.mockitoScala % Test
```
hugo-vrijswijk: Both mockitoScala and scalactic can be removed, they're not used anywhere

integration-tests/project/plugins.sbt
```diff
@@ -0,0 +1,8 @@
+resolvers += Resolver.sonatypeRepo("snapshots")
+resolvers += Resolver.sonatypeRepo("releases")
+
+addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.5.1")
+addSbtPlugin("com.geirsson" % "sbt-scalafmt" % "1.5.1")
+addSbtPlugin("org.lyranthe.sbt" % "partial-unification" % "1.1.2")
+
+addSbtPlugin("io.stryker-mutator" % "sbt-stryker4s" % "0.2.1")
```
hugo-vrijswijk: To set the version to be tested, we could do something similar to scripted sbt tests here. Setting the version as a system property (`sbt stryker -Dstryker.version=$version`)

integration-tests/src/main/scala/io/stryker4s/repository/OrderRepository.scala
```diff
@@ -0,0 +1,30 @@
...
+  def apply[F[_]: Applicative](): OrderRepository[F] = new OrderRepository[F]()
+}
+
+class OrderRepository[F[_]: Applicative] {
+
+  private val orders = new TrieMap[Int, Order]
+
+  def create(order: Order): F[Order] = {
+    orders += (order.id -> order)
+    order.pure[F]
```
hugo-vrijswijk: Not related to the PR, but how can this be a pure function if it has a side-effect (adding an element to a list)?

integration-tests/src/main/scala/io/stryker4s/validator/OrderValidator.scala
```diff
@@ -0,0 +1,30 @@
...
+
+import scala.language.higherKinds
+
+object OrderValidator {
+  def apply[F[_]: Monad](): OrderValidator[F] = new OrderValidator[F]()
+}
+
+class OrderValidator[F[_]: Monad] {
+
+  def allowedToOrderDrinks(age: Int, drinks: List[Drink]): Either[OrderNotAllowed, Unit] = {
```
hugo-vrijswijk: Isn't this just a boolean with a name? 🤔

integration-tests/src/main/scala/io/stryker4s/RoboBarServer.scala
```diff
@@ -0,0 +1,39 @@
...
+import io.stryker4s.validator.OrderValidator
+import org.http4s.implicits._
+import org.http4s.server.blaze._
+import org.http4s.server.{Router, Server}
+
+import scala.language.higherKinds
+
+object RoboBarServer extends IOApp {
+
+  private[this] val root = "/stryker4s"
```
hugo-vrijswijk: Shouldn't this be robobar?
```suggestion
  private[this] val root = "/robobar"
```

hugo-vrijswijk (commented):
Pretty cool! Not all my comments are related to the PR as much to http4s and Cats. Maybe some comments would be nice.

I feel like 30 mutations is a decent amount. What type of mutations are generated mostly? I feel like we could also use this project to look at what mutations we could add in the future.

integration-tests/project/Dependencies.scala
legopiraat: Have to see if this is compatible with http4s

integration-tests/project/Dependencies.scala
legopiraat: just leaving them in there for convenience since we might need them later

integration-tests/src/main/scala/io/stryker4s/repository/OrderRepository.scala
legopiraat: I think the response is pure in the effect context but the actual writing isn't of course. But this would/could be abstracted using a database.

integration-tests/src/main/scala/io/stryker4s/validator/OrderValidator.scala
legopiraat: In a sense yes, but you have to keep in mind that we are in a functional context where we want to for comp over the results. This makes that possible.

Also, keep in mind its custom that we collect all the errors first and return the whole reason that's rejected instead of 1 by 1 (fail fast) in some cases that make sense. In this case, it doesn't make sense to fail fast in my opinion (which is fail fast at this moment), but this implementation is close to a cats "validated" approach were we validate everything first then return the errors. but since i only have one validation rule as of now i didn't come around to build that yet.

legopiraat:
> Pretty cool! Not all my comments are related to the PR as much to http4s and Cats. Maybe some comments would be nice.
> 
> I feel like 30 mutations is a decent amount. What type of mutations are generated mostly? I feel like we could also use this project to look at what mutations we could add in the future.

At this point is mainly string mutations, also because i read in some basic drinks in the drinks repository. I tend to move that to a json file to limit these mutators and get a better overview.

hugo-vrijswijk:

PR #167, opened 2019-03-04
Make log conversion implicit

legopiraat:
#### What it does

Convert loglevel from sbt to log4j


runners/sbt/src/main/scala/stryker4s/sbt/Stryker4sPlugin.scala
```diff
@@ -42,17 +42,16 @@ object Stryker4sPlugin extends AutoPlugin {
...
-
   private def setStrykerLogLevel(level: Level.Value): Unit = {
-    val log4jLevel: org.apache.logging.log4j.Level = sbtLogLevelToLog4j(level)
-    Configurator.setRootLevel(log4jLevel)
-    LoggerContext.getContext(false).getRootLogger.info(s"Set stryker logging level to $log4jLevel")
+    Configurator.setRootLevel(level)
+    LoggerContext.getContext(false).getRootLogger.info(s"Set stryker4s logging level to $level")
   }
+
+  private[this] implicit def toLog4jLogLevel(level: Level.Value): Log4jLevel =
```
hugo-vrijswijk: I feel like this makes more sense in a separate file, similar to the other implicit conversions

legopiraat: I could do that just changed the method to an implicit so.. 😅

PR #168, opened 2019-03-04
Unify runner interface

legopiraat:
#### What it does

Moved the process runner to the actual runner module

It's not complete yet might want to do some more stuff in here


core/src/test/scala/stryker4s/Stryker4sTest.scala
```diff
@@ -17,15 +18,24 @@ import scala.util.Success
 
 class Stryker4sTest extends Stryker4sSuite with LogMatchers {
 
+  class TestMutantRunner(sourceCollector: SourceCollector)(implicit config: Config)
```
hugo-vrijswijk: Why not create a mock MutantRunner? That way you can (possibly) also test if it is being called with the correct values

docs/CONFIGURATION.md
```diff
@@ -85,6 +80,16 @@ Specify the thresholds for mutation scores.
...
 Setting `break=0` (default value) ensures that the build will never fail.
 
+## Process runner config
+#### test-runner-command
+
+**Config file:** `test-runner-command: { command: "sbt", args: "test" }`    
+**Mandatory:** Yes  
+**Description:**  
+With `test-runner-command` you specify how stryker4s can invoke the test runner. 
+Examples would be `sbt test`, `mvn test` or any other command to run your tests, including any parameters your tests might need.
```
hugo-vrijswijk: We should mention that the preferred way is through a plugin, both for performance and ease of use reasons. But should the project not be supported, this also works (but will be slower). And also make very clear that this only applies to running with the process-runner, not the plugins.

runners/command-runner/src/main/scala/stryker4s/run/Stryker4sCommandRunner.scala
```diff
@@ -1,14 +1,22 @@
...
 import stryker4s.mutants.applymutants.ActiveMutationContext.ActiveMutationContext
 import stryker4s.mutants.findmutants.SourceCollector
-import stryker4s.run.process.{Command, ProcessMutantRunner, ProcessRunner}
+import stryker4s.run.process.ProcessRunner
 import stryker4s.run.threshold.ErrorStatus
 
 object Stryker4sCommandRunner extends App with Stryker4sRunner {
 
+  private[this] val processRunnerConfig: ProcessRunnerConfig =
+    ConfigReader.readConfig[ProcessRunnerConfig]() match {
```
hugo-vrijswijk: I'm missing the tests for this that were removed in the core project

docs/CONFIGURATION.md
```diff
@@ -85,6 +80,16 @@ Specify the thresholds for mutation scores.
 
 Setting `break=0` (default value) ensures that the build will never fail.
 
+## Process runner config
+#### test-runner-command
+
+**Config file:** `test-runner-command: { command: "sbt", args: "test" }`
```
hugo-vrijswijk: As not to have *too* big of a breaking change, and to keep in line with Stryker. I think we should call this `command-runner`, or `test-runner`

hugo-vrijswijk:
The name of this PR confuses me. When I opened it, I thought it would be on the `MutantRunner` abstract class and have some refactorings there. Instead it is more about the `test-runner` config and `ProcessRunner` refactoring (which are good of course)

project/Settings.scala
```diff
@@ -17,7 +17,8 @@ object Settings {
     "-Xfatal-warnings", // Fail the compilation if there are any warnings.
     "-Xlint:doc-detached", // A Scaladoc comment appears to be detached from its element.
     "-Ywarn-infer-any", // Warn when a type argument is inferred to be `Any`.
-    "-Ywarn-dead-code" // Warn when dead code is identified.
+    "-Ywarn-dead-code", // Warn when dead code is identified.
+    "-Xmacro-settings:materialize-derivations"
```
hugo-vrijswijk: What is this used for?

PR #170, opened 2019-03-18
Update to Scalameta 4.1.5

### LINKED ISSUE #148: Mutating Scala code yields code that doesn't compile
Note sure if I should file an issue about this, or if this is expected behavior...

### Summary

Running Stryker4S on [my application](https://github.com/mthmulders/hyperion) mutates some code into a non-compiling state.

### Stryker4s config

```hocon
stryker4s {
  base-dir = "app"
  log-level = debug
  mutate = [
    "src/main/scala/**/*.scala"
  ]
  reporters = [
    console
  ]
}
```

The original code is
```scala
def retrieveMeterReading(date: LocalDate): Future[Option[HistoricalMeterReading]] = {
  val query = meterReadings.filter(_.recordDate === date)
  db.run(query.result) map (_.headOption)
}
```

and it gets mutated into
```scala
def retrieveMeterReading(date: LocalDate): Future[Option[HistoricalMeterReading]] = {
  val query = sys.props.get("ACTIVE_MUTATION").orElse(sys.env.get("ACTIVE_MUTATION")) match {
    case Some("26") =>
      meterReadings.filterNot(_.recordDate === date)
    case _ =>
      meterReadings.filter(_.recordDate === date)
  }
  db.run(query.result) map _.headOption
}
```

The complete source file is [here](https://github.com/mthmulders/hyperion/blob/master/app/src/main/scala/hyperion/database/MeterReadingDAO.scala).

### Stryker4s environment

```
stryker4s 0.1.0+19-8c5511eb-SNAPSHOT
```

### Your Environment
| Scala version        | 2.12.8 |
| Build tool & version | SBT 1.2.8 | 
| Operating System     | macOS 10.14.2  |
| Java Runtime | 1.8.0_202 |

hugo-vrijswijk:
#### What it does

Scalameta 4.1.5 provides some fixes for syntax printing of trees for issues that we ran into.

### DIFF
--- a/core/src/test/scala/stryker4s/mutants/MutatorTest.scala
+++ b/core/src/test/scala/stryker4s/mutants/MutatorTest.scala
@@ -39,9 +39,7 @@ class MutatorTest extends Stryker4sSuite with TreeEquality with LogMatchers {
                        |    case Some("3") =>
                        |      s""
                        |    case _ =>
-                       |      s"${{
-                       |        bar
-                       |      }}foo"
+                       |      s"${bar}foo"
                        |  }
                        |}""".stripMargin.parse[Source].get
       result.loneElement.tree should equal(expected)
--- a/core/src/test/scala/stryker4s/mutants/MutatorTest.scala
+++ b/core/src/test/scala/stryker4s/mutants/MutatorTest.scala
@@ -151,18 +151,4 @@ class MutatorTest extends Stryker4sSuite with TreeEquality with LogMatchers {
       "If this is not intended, please check your configuration and try again." should not be loggedAsInfo
     }
   }
-
-  describe("string interpolation") {
-
-    it("checks if the Scalameta workaround is still needed") {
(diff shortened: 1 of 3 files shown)


core/src/test/scala/stryker4s/mutants/MutatorTest.scala
```diff
@@ -151,18 +151,4 @@ class MutatorTest extends Stryker4sSuite with TreeEquality with LogMatchers {
       "If this is not intended, please check your configuration and try again." should not be loggedAsInfo
     }
   }
-
-  describe("string interpolation") {
-
-    it("checks if the Scalameta workaround is still needed") {
```
hugo-vrijswijk: We can keep this test, but I am not sure about it. We did need it before to notify us if the workaround was still needed. But I think if we keep it then we are just testing Scalameta (and that already has a lot of tests)

PR #171, opened 2019-03-20
Add vscode settings and suggested extensions

hugo-vrijswijk:
#### What it does

Adds a format-on-save setting for VS Code. And adds a suggested extension of [Metals](https://marketplace.visualstudio.com/items?itemName=scalameta.metals), which is needed for formatting, and acts as a [LSP](https://langserver.org/) for Scala.


hugo-vrijswijk:
Merged into master

PR #172, opened 2019-03-27
Write HTML report file

hugo-vrijswijk:
#### What it does

Writes the html report to file when the HTMLReporter is called. Also added tests.

When this PR is merged, I'd just like to update some documentation and then I think the `feat-html-reporter` branch can be merged to master.

#### How it works

I've also added a `FileIO` trait that abstracts file IO away (duh) for testing purposes. Idea is to later move other file IO (like in the `MutantRunner`) to this trait as well, but that felt a bit out of scope for this PR.

### DIFF
--- a/core/src/main/scala/stryker4s/files/fileIO.scala
+++ b/core/src/main/scala/stryker4s/files/fileIO.scala
@@ -1,0 +1,17 @@
+package stryker4s.files
+import better.files.{File, Resource}
+
+trait FileIO {
+  def readResource(resource: String): String
+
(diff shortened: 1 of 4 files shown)


Wmaarts:
I actually did a quick-and-dirty implementation of this a month ago in the [html-report-write-to-file branch](https://github.com/stryker-mutator/stryker4s/commit/ebb9a7bf95fe53c03e7830daabb28294a5c51427). But your implementation looks cleaner (at first glance) and isn't outdated. :p

PR #194, opened 2019-05-08
Write mutation-test-elements.js to separate file

hugo-vrijswijk:
Also log html file location to info and clean up tests a bit.

There were some issues with encoding when the mutation-test-elements.js file was an inlined script in the html. To remove any possible problems, it's now written to a separate file. This means any encoding problems can never come from us (we just copy the file from resources to disk now, nothing else)


core/src/test/scala/stryker4s/report/HtmlReporterTest.scala
```diff
@@ -45,57 +47,75 @@ class HtmlReporterTest extends Stryker4sSuite with MockitoSugar with ArgumentMat
...
+
+      val elementsCaptor = ArgCaptor[File]
+      verify(mockFileIO).createAndWriteFromResource(elementsCaptor,
+                                                    eqTo("mutation-testing-elements/mutation-test-elements.js"))
+
+      elementsCaptor.value.pathAsString should fullyMatch regex stryker4sReportFolderRegex
+      elementsCaptor.value.name equals "mutation-test-elements.js"
     }
 
     it("should debug log a message") {
```
Wmaarts: should be `"should log a message"`. :)
I agree with changing it to loglevel INFO though.

PR #197, opened 2019-05-10
Only delete stryker4s dependencies from cache

hugo-vrijswijk:
Prevents `mutation-testing-elements` from being re-downloaded each time


hugo-vrijswijk:
mutation-testing-elements is now added to cache, see [logs](https://travis-ci.org/stryker-mutator/stryker4s/jobs/530742309#L1415)

Issue #208, opened 2019-06-02
Support Stryker dashboard


leobenkel:
### Summary

I was wondering is there is any plans to have it compatible with platform like https://coveralls.io/ ? 

Thanks

hugo-vrijswijk:
Thanks for creating this issue! We don't support coveralls, as the metrics needed are a little different. However, Stryker does have a [dashboard ](https://dashboard.stryker-mutator.io/) where you can create a badge for your README after CI jobs. In the future, we also want to add support for showing a HTML report on it. Currently, Stryker4s does not support the dashboard yet. 

Is this what you are looking for? If so, I will rename this issue to supporting the dashboard

leobenkel:
So you are saying that from a Scala/SBT project, I am `NOT` able to use the dashboard today ?

leobenkel:
I just signed up to the platform and it seems that I can use it the same way Coveralls works. Just have to submit the result after the coverage test using the API key. 

But there is no SBT command to upload automatically. Is that the missing link ?

hugo-vrijswijk:
Yes. Currently the dashboard only works with Stryker for JavaScript. So not for sbt (or maven) projects. When we support it, it'll probably be integrated into the `stryker` task and upload it on CI environments after a run. Similar to Stryker for JavaScript, instead of as a separate task

leobenkel:
There is no way to do a curl push of the report ? It the only issue is the missing push of the result, there might be a hacky way to do it for now ? Or is it a deeper format issue ? 
Thanks !

hugo-vrijswijk:
Hmm, it should be possible, I guess? I imagine Stryker just does a POST to the dashboard. It's not an 'official' way, but you might be able to find something in the Stryker source code

leobenkel:
So it works when you do:

```
curl -X POST \
  https://dashboard.stryker-mutator.io/api/reports \
  -H 'Content-Type: application/json' \
  -d '{
    "apiKey": "[API_KEY]",
    "repositorySlug": "github.com/[USER]/[REPO_NAME]",
    "branch": "[BRANCH]",
    "mutationScore": [MUTATION_SCORE]
}'
```

leobenkel:
Now I need to find a way to read the actual percent...

leobenkel:
Is there a reporter that would dump a json file maybe where I could read the final score?

hugo-vrijswijk:
[...]

PR #209, opened 2019-06-03
Add a Json reporter

leobenkel:
https://github.com/stryker-mutator/stryker4s/issues/208

#### What it does

It add a json output to be parse by bash to send curl to dashboard 

#### How it works

Just add `json` as reporter and a file will be created at 
```
config.baseDir / s"target/stryker4s-result-${System.currentTimeMillis()}" / "result.json"
```

The file will contains `mutationScore` and `durationSeconds`.

### DIFF
--- a/core/src/main/scala/stryker4s/config/ReporterType.scala
+++ b/core/src/main/scala/stryker4s/config/ReporterType.scala
@@ -11,3 +11,7 @@ case object ConsoleReporterType extends ReporterType {
 case object HtmlReporterType extends ReporterType {
   override val name: String = "html"
 }
+
+case object JsonReporterType extends ReporterType {
+  override val name: String = "json"
+}
--- a/core/src/main/scala/stryker4s/model/MutantRunResult.scala
+++ b/core/src/main/scala/stryker4s/model/MutantRunResult.scala
@@ -4,7 +4,9 @@ import java.nio.file.Path
 
 import scala.concurrent.duration.Duration
 
-case class MutantRunResults(results: Iterable[MutantRunResult], mutationScore: Double, duration: Duration)
+case class MutantRunResults(results: Iterable[MutantRunResult], mutationScore: Double, duration: Duration) {
+  @transient val timestamp: Long = System.currentTimeMillis()
+}
 
 /** The base result type of a mutant run.
   * Extends Product with Serializable to clean up the type signature, as all subtypes are case classes
--- a/core/src/main/scala/stryker4s/report/HtmlReporter.scala
+++ b/core/src/main/scala/stryker4s/report/HtmlReporter.scala
@@ -48,7 +48,7 @@ class HtmlReporter(fileIO: FileIO)(implicit config: Config)
   }
 
   override def reportRunFinished(runResults: MutantRunResults): Unit = {
-    val targetLocation = config.baseDir / s"target/stryker4s-report-${System.currentTimeMillis()}"
+    val targetLocation = config.baseDir / s"target/stryker4s-report-${runResults.timestamp}"
 
     val mutationTestElementsLocation = targetLocation / mutationTestElementsName
     val indexLocation = targetLocation / "index.html"
--- a/core/src/main/scala/stryker4s/report/JsonReporter.scala
+++ b/core/src/main/scala/stryker4s/report/JsonReporter.scala
@@ -1,0 +1,33 @@
+package stryker4s.report
+
+import better.files.File
+import grizzled.slf4j.Logging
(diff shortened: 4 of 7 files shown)


hugo-vrijswijk:
Thanks for your PR! How do you envision the future of this reporter? If you see it as a temporary thing before we support the dashboard, a log warning to let the user know that this is a temporary thing would be useful, I think (and in documentation as well). If you want to have a more general JSON reporter for other usages, i think it should output the JSON report in the same format as the HTML reporter. As that is a framework-agnostic way of outputting mutation results.  

What do you think?

core/src/main/scala/stryker4s/report/JsonScoreReporter.scala
```diff
@@ -0,0 +1,55 @@
...
+
+    def toJson: String = {
+      import io.circe.generic.auto._
+      import io.circe.syntax._
+      this.asJson.noSpaces
+    }
+  }
+
+  override def reportRunFinished(runResults: MutantRunResults): Unit = {
+    val targetLocation = config.baseDir / s"target/stryker4s-result-${System.currentTimeMillis()}"
```
hugo-vrijswijk: Keep in mind this means the html and JSON are written to different folders, as they will be called after each other (different currentTimeMillis)

leobenkel:
> Thanks for your PR! How do you envision the future of this reporter? If you see it as a temporary thing before we support the dashboard, a log warning to let the user know that this is a temporary thing would be useful, I think (and in documentation as well). If you want to have a more general JSON reporter for other usages, i think it should output the JSON report in the same format as the HTML reporter. As that is a framework-agnostic way of outputting mutation results.
> 
> What do you think?

I think it would be great, but i am not sure what format the html reporter is using. Could you point me to this part of the code? Thanks

core/src/main/scala/stryker4s/report/JsonScoreReporter.scala
leobenkel: Fixed in the coming push

leobenkel:
* I update the reporter to follow the same pattern as the HTML one.
* Copied the unit test from the HTML reporter to test the JSON one.
* Fix the timestamp discrepancy
* The output folder is the same for both test now

hugo-vrijswijk:
@leobenkel I am still unsure about the JSON reporter in itself. If I am correct, your use case is to upload a score to the dashboard. Reading the score from a JSON and then curl'ing it seems like the wrong way to go, as I think Stryker4s could do this in its own dashboard-reporter (in the future). That way you don't need to write any scripts of your own reading the report and using curl. 

I see two possible use cases for a JSON reporter:
- As a way to read the score and POST curl to the dashboard
- As a more general way for other possible use cases

So my question is if this assumption is correct, and which you want to use the reporter for? For the first I think this PR is great as is to merge and we could create an issue to rewrite it to a dashboard reporter at a later point. For the second I think we should write the general JSON schema to a file (see my comments above).

leobenkel:
I'm going to fix your comment today.

That would be amazing to send directly to the dashboard, however, the api require the repo slug, the branch and the api key. I am not how to get all of those from within the code. However, it is trivial from bash / travis environment. That's why I was thinking that printing the score in a file then reading it from bash then curl was the fastest way to get to a working dashboard uploader solution.

I don't mind using the case class that the html reporter is using but why not having the score already calculated in there ? That will add extra complexity to the consumer of the report files. That might also open up to errors if the calculation of the score is implemented differently in all the consumers. 

I can:

* Extends the html reporter class to include the score ?
* Move the timestamp to the result class
* Open a ticket to implement a direct dashboard reporter for the future
* I am open to suggestion on how to fetch the missing element to POST to the dashboard directly.

leobenkel:
I updated to comply with the result format but I really dont see the point anymore since the output json doesnt contain anything useful: 

https://github.com/stryker-mutator/stryker4s/pull/209/files#diff-a0712e58c2b9ab9884cd7f2586e65a35R33

leobenkel:
I can make a first shot at it, make a skeleton and you can take it over and add the passing of the env var to the reporter. 
Let me try to do that today.

hugo-vrijswijk:
Awesome! I've already made a small start on it in the [`dashboard-reporter` branch](https://github.com/stryker-mutator/stryker4s/compare/dashboard-reporter)

leobenkel:
ah oups @hugo-vrijswijk https://github.com/stryker-mutator/stryker4s/pull/210

hugo-vrijswijk:
Haha should've checked my email I guess

PR #210, opened 2019-06-06
Add Dashboard reporter

leobenkel:
#### What it does

Add a dashboard reporter

#### How it works

hit the dashboard api reporter URL and post the results when test is completed

### DIFF
--- a/core/src/main/scala/stryker4s/config/ReporterType.scala
+++ b/core/src/main/scala/stryker4s/config/ReporterType.scala
@@ -15,3 +15,7 @@ case object HtmlReporterType extends ReporterType {
 case object JsonReporterType extends ReporterType {
   override val name: String = "json"
 }
+
+case object DashboardReporterType extends ReporterType {
+  override val name: String = "dashboard"
+}
--- a/core/src/main/scala/stryker4s/config/ReporterType.scala
+++ b/core/src/main/scala/stryker4s/config/ReporterType.scala
@@ -21,7 +21,7 @@ case object JsonReporterType extends ReporterType {
 case object DashboardReporterType extends ReporterType {
   override val name: String = "dashboard"
 
-  def unapply(reporterType: ReporterType)(implicit config: Config): Option[DashboardReporter] = reporterType match {
+  def unapply(reporterType: ReporterType): Option[DashboardReporter] = reporterType match {
     case DashboardReporterType => DashboardReporter.resolveProvider()
     case _                     => None
   }
--- a/core/src/main/scala/stryker4s/config/ReporterType.scala
+++ b/core/src/main/scala/stryker4s/config/ReporterType.scala
@@ -1,5 +1,7 @@
 package stryker4s.config
 
+import stryker4s.report.DashboardReporter
+
 sealed trait ReporterType {
   val name: String
 }
@@ -18,4 +20,9 @@ case object JsonReporterType extends ReporterType {
 
 case object DashboardReporterType extends ReporterType {
   override val name: String = "dashboard"
+
+  def unapply(reporterType: ReporterType)(implicit config: Config): Option[DashboardReporter] = reporterType match {
+    case DashboardReporterType => DashboardReporter.resolveProvider()
+    case _                     => None
+  }
 }
--- a/core/src/main/scala/stryker4s/http/webIO.scala
+++ b/core/src/main/scala/stryker4s/http/webIO.scala
@@ -4,46 +4,21 @@ trait WebIO {
   def postRequest(url: String, content: String): WebResponse
 }
 
-case class WebResponse(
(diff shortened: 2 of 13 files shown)


leobenkel:
@hugo-vrijswijk you can take over to pass the value down to the reporter

hugo-vrijswijk:
Awesome! Thanks a lot. I'll take a closer look at it tomorrow (EU time here)

core/src/main/scala/stryker4s/report/DashboardReporter.scala
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  private val DashboardRootURL: String = "https://dashboard.stryker-mutator.io"
+  private val DashboardURL: String = s"$DashboardRootURL/api/reports"
+
+  def buildScoreResult(input: MutantRunResults): StrykerDashboardReport = {
+    StrykerDashboardReport(
+      "PLACEHOLDER_API_KEY", // TODO: Get apikey
+      "PLACEHOLDER_REPO_SLUG", // TODO: Get repo slug
+      "PLACEHOLDER_BRANCH_NAME", // TODO: Get branch name
```
leobenkel: @hugo-vrijswijk you just have to fill those values and I think we are done.

hugo-vrijswijk:
@leobenkel I've made some changes, could you do a review of them?

core/src/main/scala/stryker4s/config/ReporterType.scala
```diff
@@ -18,4 +20,9 @@ case object JsonReporterType extends ReporterType {
 
 case object DashboardReporterType extends ReporterType {
   override val name: String = "dashboard"
+
+  def unapply(reporterType: ReporterType)(implicit config: Config): Option[DashboardReporter] = reporterType match {
```
leobenkel: missing returned type. It speed up compilation

core/src/main/scala/stryker4s/report/DashboardReporter.scala
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  def writeReportToDashboard(url: String, report: StrykerDashboardReport): WebResponse = {
+    webIO.postRequest(url, report.toJson)
+  }
+
+  override def reportRunFinished(runResults: MutantRunResults): Unit = {
+    val response = writeReportToDashboard(dashboardURL, buildScoreResult(runResults))
+
+    if (response.httpCode == 201) {
```
hugo-vrijswijk: Should be something like "is in OK range (2xx)"

core/src/main/scala/stryker4s/http/webIO.scala
```diff
@@ -0,0 +1,24 @@
+package stryker4s.http
+
+trait WebIO {
+  def postRequest(url: String, content: String): WebResponse
+}
+
+case class WebResponse(httpCode: Int, responseBody: String)
+
+object RealHttp extends WebIO {
```
hugo-vrijswijk: Can we rename this to HttpClient it feels more natural of a name to me.

core/src/main/scala/stryker4s/report/DashboardReporter.scala
hugo-vrijswijk: Fixed, now uses scalaj-http response which has `isSuccess`

legopiraat:
Thank you @leobenkel!

nicojs:
Awesome new feature, thanks! 👍

leobenkel:
What version will it be part of ?

hugo-vrijswijk:
You should already be able to use the snapshot. I'll make a release (0.5.0) soon (no promises)

leobenkel:
> You should already be able to use the snapshot. I'll make a release (0.5.0) either soon (no promises) or tomorrow

cool thanks. I'll wait for `0.5.0` .

hugo-vrijswijk:
@leobenkel I just [started a new release process](https://travis-ci.org/stryker-mutator/stryker4s/builds/544417659). Once that finishes and syncs to Maven Central it should be available. I'll write release notes tomorrow

leobenkel:
@hugo-vrijswijk just wanted to let you know that the dashboard worked ! https://travis-ci.org/leobenkel/safety_plugin/builds/547924432 It was submitted successfully!
The badge got updated: https://badge.stryker-mutator.io/github.com/leobenkel/safety_plugin/master 

It was pretty straightforward to use ! The only issue I had was the size of the log exploded in Travis face so I had to change the log level to Error only to be able to fully run the entire thing.

hugo-vrijswijk:
Awesome to hear! I've noticed that on CI tests of mutation runs will also be logged, even though they should only be logged on debug. So I'll have to look into that. Thanks for the feedback!

PR #222, opened 2019-06-15
Fix dashboard reporter

### LINKED ISSUE #221: Dashboard reporter does not work
### Summary

The dashboard reporter currently does not work. This is because the POST request we create does not contain the proper information. The `repositorySlug` that is sent just has the `user/repo`, instead it should be `github.com/user/repo`. When this happens, a `200` with a body of `null` is sent back, instead of a 201. So the fix should also check for a 201.

### Stryker4s config

```hocon
stryker4s {
  reporters = [ "dashboard"]
}
```

### Stryker4s environment

```
stryker4s 0.5.1
```

### Your Environment
| Build tool & version |      Travis      |

hugo-vrijswijk:
#### What it does

Dashboard reporter should now work

#### How it works

- Prepend `github.com/` to repositorySlug
- Always check for 201 instead of 2xx
- Fix and add to tests


hugo-vrijswijk:
@legopiraat approves, but is not logged in

PR #283, opened 2019-10-17
Improve logging during config load

### LINKED ISSUE #269: Log warnings for deprecated configuration
There's some configuration that has been changed/removed since initial release. Such as the `log-level` in [v0.3.1](https://github.com/stryker-mutator/stryker4s/releases/tag/v0.3.1), `command-runner` in [v0.4.0](https://github.com/stryker-mutator/stryker4s/releases/tag/v0.4.0).

In order to let the user know that these settings no longer work, or for future deprecated configuration, we should add `WARN` logging of changed configuration, showing both what setting is changed, and what the new configuration looks like (possibly with a link to the config documentation). We could also log when a configuration setting is equal to the default

The `Config` case class should be kept clean of these removed settings.

As of a recent Pureconfig version, it is possible to [load multiple case classes from the same config 'source'](https://pureconfig.github.io/docs/loading-a-config.html#loading-a-config-in-a-path). That could be useful for us in order to only read from the config file once.

scubamaggo:
#### How it works

The test for non-existing keys is done by using the PureConfig configuration to fail config load when encountering unknown keys. We then log these failures and attempt to load the config again with a less strict setting, allowing the existence of unknown keys.


hugo-vrijswijk:
Thank you very much for your PR! Sorry I haven't been able to take a look. I hope to be able to do so tonight. I notice the PR is still a draft. Could you explain what you still think needs to be done? 

Regarding your question about the ProcessRunner on Gitter: I personally wouldn't put too much effort into getting the perfect experience for it. The ProcessRunner is not used much and we provide plugins for the two most-used Scala build tools.

I also noticed that although all tests were passing, formatting failed. You can run `./bin/scalafmt` to fix any formatting.

scubamaggo:
No problem, I haven't had time to do much in the meantime anyways.
This PR is in draft mode, because I still have/had the following on my to-do list, ordered by (perceived) importance:

1. Discuss with the introduced Either syntax with the maintainers. It's basically a copy-paste from cats-core. One option would be to add a dependency. Another would be to keep the code as is, but move it to a more appropriate location.
2. Do the logging for the ProcessRunner config. I would omit this after your comments.

1.

core/src/main/scala/stryker4s/config/ConfigReader.scala
```diff
@@ -4,45 +4,152 @@ import java.io.FileNotFoundException
...
+    * provide a convenient way to attempt to read a config with various [[Derivation]]s,
+    * depending on the [[ConfigReaderFailures]] that was returned from the last attempt.
+    * In addition to that, some logging is also enforced during the reading process.
+    * If not for those points, simply using the returned [[Either]] of the [[PureConfigReader]]
+    * would be sufficient as well, since the exposed API here is basically a subset of [[Either]]s.
+    * @param file the [[File]] from which the config is to be read.
+    * @param derivation the [[Derivation]] that is used to configure the [[PureConfigReader]].
+    * @tparam T the type of the config that is to be read.
+    */
+  private class Reader[T] private (file: File)(implicit derivation: Derivation[PureConfigReader[T]]) {
```
hugo-vrijswijk: I like this construction of a builder-like reader with composable behaviour a lot! However, I feel like the mutability can be avoided. The functions could be in the constructor, with the empty PF as defaults. Each function that returns a `Reader` could return a new instance of a `Reader` with the new functions composed.

Lastly, this class is getting quite big (in part due to the comments 👍). Could you move it to a separate file and extract any config-reader specific stuff from it (like logging), just to separate the concerns and make testing easier?

core/src/main/scala/stryker4s/config/ConfigReader.scala
```diff
@@ -4,45 +4,152 @@ import java.io.FileNotFoundException
...
+      * The names of the unknown keys are logged.
+      */
+    def onUnknownKey: PartialFunction[ConfigReaderFailures, Derivation[PureConfigReader[Config]]] = {
+      case ConfigReaderFailures(ConvertFailure(UnknownKey(key), _, _), failures) =>
+        val unknownKeys = key :: failures.collect {
+          case ConvertFailure(UnknownKey(k), _, _) => k
+        }
+
+        warn(
+          s"The following configuration keys are not used: ${unknownKeys.mkString(", ")}.\n" +
```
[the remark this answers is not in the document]
scubamaggo: I could change the logged text accordingly, i.e. sth like

> The following configurations keys are not used, maybe the stem from an older stryker4s version: ...

I don't think we should be explicitly checking for certain keys that once were used in the `Config` as this requires us to hold a reference to every key that once was.

core/src/main/scala/stryker4s/config/ConfigReader.scala
scubamaggo: I will remove the mutable instance. I originally placed it here since we are in very controlled, safe space and thus it felt ok to avoid to cost of instantiating a new object for every api-call but since reading the config is not on any hot-path whatsoever, I agree that this is unnecessary.

I can also move the class, but I'm not sure about pulling the logging out of it. One of the reason for it's existence was to force some level of common logging when reading configs of different types. 
Taking another look at the logging though made me realise that the info that is logged on a successful read is wrong. The current text is only correct for the default file-path. 
How about I change the logging such that the path of the config that was read is taken from the input to the class?

scubamaggo: I am not sure anymore about moving the class. What should be its scope? It's only functionality is to configure the reading from config. That's exactly what `ConfigReader` does. If I remove it from the `ConfigReader`, that object will only contain a call to that class. 
Isn't it good that we are testing the `ConfigReader` and not its implementation details?

hugo-vrijswijk: > If I remove it from the ConfigReader, that object will only contain a call to that class.
Isn't it good that we are testing the ConfigReader and not its implementation details?

Taking another looking at it and reading your comment, I agree! It makes perfect sense for it to be an inner class 👍

hugo-vrijswijk:
I've added a suggestion to slightly change the wording of the logged warning message. If you add that I will approve and merge this 😁👍

hugo-vrijswijk:
Thank you for this PR!

PR #407, opened 2020-03-20
Add cache to github actions

hugo-vrijswijk:


hugo-vrijswijk:
Doesn't seem to have any speedups, while adding a **lot** of verbosity. Closing for now.

PR #461, opened 2020-05-15
Add test-filter to Config

CharlesD91:
#### What it does

Adds `test-filter` to `Config`

#### How it works

It restricts the tests that are run for every mutation, so that we can ignore heavy `integration` tests and focus on `unit` ones

### DIFF
--- a/core/src/test/scala/stryker4s/config/TestFilterTest.scala
+++ b/core/src/test/scala/stryker4s/config/TestFilterTest.scala
@@ -1,0 +1,32 @@
+package stryker4s.config
+
+import stryker4s.testutil.Stryker4sSuite
+
+class TestFilterTest extends Stryker4sSuite {
+  describe("filter with default") {
+    implicit val conf: Config = Config.default
+    it("should work") {
+      val testFilter = new TestFilter
+      assert(testFilter.filter("stryker4s.extension.FileExtensionsTest"))
+    }
+  }
+
+  describe("filter with custom") {
+    implicit val conf: Config = Config.default.copy(testFilter = Seq("stryker4s.extension.*"))
+    it("should work") {
+      val testFilter = new TestFilter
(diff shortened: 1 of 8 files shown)


core/src/test/scala/stryker4s/config/TestFilterTest.scala
```diff
@@ -0,0 +1,32 @@
+package stryker4s.config
+
+import stryker4s.testutil.Stryker4sSuite
+
+class TestFilterTest extends Stryker4sSuite {
+  describe("filter with default") {
+    implicit val conf: Config = Config.default
+    it("should work") {
```
hugo-vrijswijk, line 8: The 'should work' is a little duplicate three times. Perhaps a single `describe` of the function you are testing, with an `it` of the changed system under test description would be clearer?

CharlesD91: I fixed this

CharlesD91:
> Thank you for your PR! Overall the code looks good, I just have a small comment. Could you also add some documentation about the new config option?

I updated the doc. Also I did a small change on how the `TestFilter` is used in the `SbtMutantRunner` settings: if the `test-filter` wasn't set in the config, then nothing related to it will happen in the code (better sandboxing of the different logics in case of a bug)

cbrunnkvist:
I was just about to open a bug report for v0.7.3 about this not working correctly, when I by chance discovered this PR and that it was closed less than a week ago. 👍 

I had perhaps naively assumed that the `test-filter` option was already in all recent releases. Perhaps this means it is time to start updating _"added in version..."_ to the CONFIGURATION.md items. 😓

hugo-vrijswijk:
@cbrunnkvist Excellent idea! Would be helpful for a lot of people. In general, our documentation always reflects the latest version. But feel free to open a PR!

Issue #412, opened 2020-03-31
Run initial compile before initial test run


hugo-vrijswijk:
### Current situation

After mutating and writing code, Stryker4s does an initial test run to make sure tests pass. When they do, Stryker4s continues. If they don't Stryker4s stops with an error.

### The problem

It's possible Stryker4s mutates code in an invalid way. For example, creating compile errors or simple syntax errors for whatever reason.

When code is mutated in an invalid way, the user only gets the error that the initial test run failed, nothing more. This does not give a lot of information about what is probably going on and what they could do to fix it (before Stryker4s handles it itself in a better way).

### Desired situation

We could do an initial compile of mutated code **before** the initial test run. If this fails, better communication can be given to the user about any compile errors if they occur. 

We can ask the user to create an issue with an example, as mutated code _should_ always be valid

WaDelma:
I encountered this when trying `stryker4s` out on a project, getting the following error.
```
value nonEmpty is not a member of com.twitter.finagle.Path
```
The message `Initial test run failed. Please make sure your tests pass before running Stryker4s.` is really confusing because the tests definitely passed when running with `mvn test`.

WaDelma:
After ignoring `MethodExpression` I now get bunch of `not found: value prefix` errors. Is there way to see the generated code to inspect what is going on?

Issue #312, opened 2019-12-02
Feature proposal: time-out detection


mthmulders:
Due to mutated Scala code, it is possible that the code under test contains an endless loop. Right now, Stryker4s would just wait for the test to finish.

My suggestion is to build a timeout detection (just like [Stryker for Javascript has](https://github.com/stryker-mutator/stryker/tree/master/packages/core#timeoutMS)). It should detect that the test runs way too long, and then change the mutant status to "TimedOut" or something like that.

I'd love to play around and see if I can build it, but I have very limited time at this moment. Let's keep this issue for discussing whether this is needed, and how it should work.

hugo-vrijswijk:
Closed in #573

Issue #16, opened 2018-06-22
Sort tests on speed


hugo-vrijswijk:
Run the fastest tests first for each mutation. This way (hopefully) the tests cause a killed mutant sooner

nicojs:
This seems like a complicated thing to do. Maybe impact analysis and multi-threaded runs should go first?

hugo-vrijswijk:
Definitely. I've created a [Waffle.io board](https://waffle.io/stryker-mutator/stryker4s) and sorted the issues on what I think should go first

nicojs:
It would if you have some integration tests intermingled with your unit tests 🤷‍♀️

Issue #10, opened 2018-06-22
Run mutants multithreaded


hugo-vrijswijk:
To improve performance, mutants can be ran multithreaded. Should be dependent on the number of cores on the machine or set via a configuration option

legopiraat:
We could make it an optional configuration field. 

By default we could use something like `int cores = Runtime.getRuntime().availableProcessors() - 1`

the minus one is on purpose so we dont fully kill the machine's performance.

hugo-vrijswijk:
I agree with both of those. Having cores - 1 also leaves room for an orchestrator of some kind.

We should probably also let the user know how many test runners are started, and based on what (config or processor count), same as Stryker

hugo-vrijswijk:
It looks like SBT [runs all tests in parallel by default](https://www.scala-sbt.org/1.x/docs/Testing.html#Disable+Parallel+Execution+of+Tests). Not sure how Maven does this, but that could mean this might be a slightly lower priority for now. There might not be that big of a performance boost from running in parallel (or it might have a negative impact on performance with lots of timeouts)

legopiraat:
Agreed for sure if we could leave the sbt process open it would be a huge gain. Am trying to come up with a way to remove sbt from the testruns in full. But it has some complications on the side which i need to investigate.

cbrunnkvist:
There are several reasons to why a project can decide _not_ run local tests in parallel... 

A more scalable approach than just local parallelism would perhaps be some method of deciding a "partition ID", so that we can use CI pipelines to construct as much parallelism as we want. This would be fairly advanced to set up, but it would be quite useful to _have_ that option: The One Thing that stops this from being even remotely possible at the moment is _the lack of separation_ between _running the mutation tests_ and _generating the human-readable report_ (i.e. akin to how [sbt-scoverage](https://github.com/scoverage/sbt-scoverage) separates `sbt coverage` from `sbt coverageReport`).

PR #492, opened 2020-06-02
Sbt process testrunner

### LINKED ISSUE #152: Improve how much control we have over mutation runs
Right now there is no control over the process when running a mutation. This means when a mutation causes an infinite loop we can't do much about it. It'd be great to have more control over the process. 

I have 2 ideas about how we could do this:
- Creating a sub-process/agent that then calls the test tasks. Should the mutation run take too long we can kill the agent and start a new one. Downside is killing the agent doesn't kill the sub-process (I think). Also means (I think) the build tool would have to start for each agent.
- Starting tests by starting a new java process with the appropriate classpath, main class, etc. This means collecting the classpath and all that first, then calling `java -cp ...`. This gives a lot of control over the process (as much as possible, really). Downside is likely a lot more work to implement, and possibly test-runner specific work (Scalatest, JUnit, all have different entry points). This is also, for example, how IntelliJ runs tests. Looking at the intellij-scala plugin a lot of test-tool specific code happens there.

A combination of the two might also be possible, with agents starting new test runs.

There is some definite uncertainty for me in this area, as I am not very familiar with either of the two solutions.

hugo-vrijswijk:
### Closes #152 

Changes the sbt-plugin to run tests in a separate process. The plugin will start a new process with the new sbt-testrunner project and the application to test on the classpath. Also adds an api project to describe messages that are sent between the main and sub-processes.

The sbt-testrunner project will programmatically start tests using the [sbt test-interface API](https://github.com/sbt/test-interface). I made a small [POC to run tests programmatically](https://github.com/hugo-vrijswijk/test-init-poc) with the interface using either ScalaTest or MUnit. Both of those frameworks (and any testing frameworks that work in sbt) implement the test-interface.

To do:

- [x] Define projects
- [x] Start new java process from the main process
  - [x] Collect test context (test names to run, classpath, etc)
  - [x] Manage process in sbt plugin (starting, ending)
  - [x] Resolve correct scala-compatible version of child-process library (~~probably using [Coursier](https://get-coursier.io/docs/api)~~ adding the library to the `libraryDependencies` in sbt)
    - [x] Explore as many scalaVersions as we can for support. Currently Scala 2.11, 2.12, 2.13 and Dotty 0.27
- [x] Send messages to subprocess (probably using sockets?)
  - [x] ~~Experiment with Java RMI~~ Not needed. Serialization works well and keeps us in charge of the process
- [x] Start listening for messages in new test-process
- [x] Setup state for new process (tests to run, initialize framework, etc)
- [x] Handle messages (start tests, end process, etc)
- [x] Report results back from sub- to main-process
- [x] ~~Think of a better name for the sub-process package~~ Simply sbt-testrunner. If we expand it to maven the package can be renamed
- [x] Cleanup
- [x] Fix/improve/add tests
- [x] Fix/investigate issue with StackoverflowError after ~80 mutants (when running on core project)
- [x] Timeout detection (hardcoded duration)
- [x] Retry mechanism for starting new processes
- [x] Pass-through java arguments

Out of scope for this PR:

- Plugins that aren't sbt. 
  - This is probably possible for the maven plugin, but I want a working MVP first. And it would likely require additional setup for junit tests or other test frameworks as those don't implement the sbt test-interface
- Setups with multiple testing frameworks. Assumption is a single test-framework. Will probably work, but will be verified after this PR.
- Detecting static mutants (in objects, for example, the code will be run before any mutations are activated, so they can never be mutated). We should ignore these in the future
- Filtering tests on coverage
- Filtering tests on speed
- Starting multiple processes for performance
- Scala.js. JVM only for now

It also opens up the way towards features like #412, #312, #16 and #10 (wow that's a big list)


project/Dependencies.scala
```diff
@@ -25,6 +26,7 @@ object Dependencies {
     val mockitoScala = "org.mockito" %% "mockito-scala-scalatest" % versions.mockitoScala % Test
   }
 
+  val testInterface = "org.scala-sbt" % "test-interface" % versions.testInterface
```
nicojs, line 29: Curious. What happens if the project under test depends on a different version of this interface?

sbt-testrunner/src/main/scala/stryker4s/sbt/testrunner/SbtTestRunner.scala
```diff
@@ -0,0 +1,49 @@
...
+    var status = acc
+    val eventHandler: EventHandler = event => {
+      status = combineStatus(status, event.status())
+    }
+    val newTasks = testTasks.flatMap(task => task.execute(eventHandler, Array.empty))
+    if (newTasks.nonEmpty) runMutation(newTasks, status)
+    else status
+  }
+
+  def combineStatus(current: Status, newStatus: Status) =
```
nicojs, line 41: This def seems awfully verbose given what it's doing 🙈

sbt/src/main/scala/stryker4s/sbt/runner/ProcessHandler.scala
```diff
@@ -0,0 +1,18 @@
+package stryker4s.sbt.runner
+
+import scala.tools.nsc.io.Socket
+import stryker4s.model.MutantRunResult
+import java.nio.file.Path
+
+class ProcessHandler() {
+  def newProcess(classpath: Seq[String]) = {
+    val socket = new Socket(new java.net.Socket("127.0.0.1", 13337))
```
nicojs, line 9: Old school. Inter process communication using sockets?

sbt/src/main/scala/stryker4s/sbt/runner/SbtMutantRunner.scala
```diff
@@ -9,13 +9,11 @@ import sbt._
...
 import stryker4s.extension.FileExtensions._
-import stryker4s.extension.exception.InitialTestRunFailedException
 import stryker4s.model._
 import stryker4s.mutants.findmutants.SourceCollector
 import stryker4s.report.Reporter
 import stryker4s.run.MutantRunner
 import stryker4s.sbt.Stryker4sMain.autoImport.stryker
-import sbt.Tests.Output
 
 class SbtMutantRunner(state: State, sourceCollector: SourceCollector, reporter: Reporter)(implicit config: Config)
```
nicojs, line 18: What is a `MutantRunner`? I'd imagine it is the `TestRunner` adapter for Stryker4s? Is it also used for the initial test run?

project/Dependencies.scala
hugo-vrijswijk: The last version of this package was published in 2013, so I would say that's a pretty stable API. Even so, I've now changed the dependency to 'Provided' (so assumed to be there). As long as there are no breaking changes used in the API it should be fine

sbt-testrunner/src/main/scala/stryker4s/sbt/testrunner/SbtTestRunner.scala
hugo-vrijswijk: Open for PR's 😊. Everything in this PR is very WIP at the moment

sbt/src/main/scala/stryker4s/sbt/runner/SbtMutantRunner.scala
hugo-vrijswijk: Yes and yes. Does this mean I've named it correctly? I'm not sure TBH. It's responsible for running tests with the correct mutation activated (there's one for sbt, maven, command-runner)

sbt/src/main/scala/stryker4s/sbt/runner/ProcessHandler.scala
hugo-vrijswijk: Yeah... If it works don't fix it? The issue is I need some way to talk to talk to a process that is:

- Two-way communication
- Not stdin. Anything can write to stdin using `println`
- Doesn't introduce new dependencies. Could cause conflicts with the system under test
- Able to serialize jvm objects through. JSON/protobuf/anything else introduces new dependencies, see point 2

So I ended up with sockets. It's very low-level, but I'm open to suggestions haha

hugo-vrijswijk:
> I have more questions than anything else 😀

Exactly what I was hoping for

PR #499, opened 2020-06-12
Fixes stryker-mutator/stryker4s#495

### LINKED ISSUE #495: Use annotations to skip mutations
### Summary
It would be cool, if I could disable some lines of code from some checking.
For example when I configure my HTTP client logging option with booleans.
```  private val client = RequestLogger(logHeaders = false, logBody = false)(ResponseLogger(logHeaders = false, logBody = true)(httpClient))```
I dont need to extract values to config. I also dont want to disable whole file from testing. One thing I need - disable this exact code from boolean inversion checking.

#### My proposal
If its possible, we could use java.lang.SuppressWarnings annotation. For example before code line put `SuppressWarnings(Array("stryker4s.extension.mutationtype.BooleanLiteral"))`
Then, engine could skip this line from checking

### PR DESCRIPTION
**Pull request by MercurieVV**
Use @SuppressWarnings annotation to skip mutations

#### What it does

This PR allow skip particular mutations in annotated context. For example if we dont need mutate exact line of code (like constant value)

#### How it works

It will iterate to root of syntax Tree checking for @SuppressWarnings annotation. If this annotation contains id of mutation (for example `@SuppressWarnings(Array("stryker4s.extension.mutationtype.StringLiteral))`), then this mutation will be skipped

### DIFF
--- a/core/src/main/scala/stryker4s/mutants/findmutants/MutantMatcher.scala
+++ b/core/src/main/scala/stryker4s/mutants/findmutants/MutantMatcher.scala
@@ -87,9 +87,8 @@ class MutantMatcher()(implicit config: Config) {
     ): Seq[Option[Mutant]] =
       ifNotInAnnotation {
         mutations
-          .filterNot(isSuppressedByAnnotation(_, original))
           .map { mutated =>
-            if (matchExcluded(mutated))
+            if (matchExcluded(mutated) || isSuppressedByAnnotation(mutated, original))
               None
             else
               Some(Mutant(ids.next, original, mutationToTerm(mutated), mutated))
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: core/src/main/scala/stryker4s/mutants/findmutants/MutantMatcher.scala [hunk 1]
```diff
@@ -84,12 +86,14 @@ class MutantMatcher()(implicit config: Config) {
...
     ): Seq[Option[Mutant]] =
       ifNotInAnnotation {
-        mutations map { mutated =>
-          if (matchExcluded(mutated))
-            None
-          else
-            Some(Mutant(ids.next, original, mutationToTerm(mutated), mutated))
-        }
+        mutations
+          .filterNot(isSuppressedByAnnotation(_, original))
```
hugo-vrijswijk, line 90: Can you move this to the `if` in the map-statement so it will count as excluded? That will probably make it easier in the future to show this as `Ignored` status in the report.

MercurieVV: Thanks for review! 100% agree with all your notes. Just pushed changes.

PR #550, opened 2020-08-10
Propagate active profiles from Maven plugin configuration to stryker runs

### LINKED ISSUE #541: Filter out blocklisted Scala compiler options for Maven plugin
Generated mutated code can cause certain compiler options to warn (unused variables, etc). This combined with `fatal-warnings` can cause compile errors after mutating code. In the sbt plugin, [those options are known](https://github.com/stryker-mutator/stryker4s/blob/master/sbt/src/main/scala/stryker4s/sbt/runner/SbtMutantRunner.scala#L24-L33) and [filtered out](https://github.com/stryker-mutator/stryker4s/blob/master/sbt/src/main/scala/stryker4s/sbt/runner/SbtMutantRunner.scala#L54) when Stryker4s runs. 

The same should be done for the Maven plugin to prevent the same issues.

For Maven, to start tests in Stryker4s the maven invoker plugin is used by building a [request](https://github.com/stryker-mutator/stryker4s/blob/master/maven/src/main/scala/stryker4s/maven/runner/MavenMutantRunner.scala#L52-L57) and [starting it](https://github.com/stryker-mutator/stryker4s/blob/master/maven/src/main/scala/stryker4s/maven/runner/MavenMutantRunner.scala#L36). The fix is probably in this area, but I am not familiar enough with how compiler options work in the Scala Maven plugin to provide more information.

jozic:
#### What it does

Propagate active profiles from stryker4s maven plugin configuration to stryker runs, which allows to change different settings specifically for stryker4s runs

### DIFF
--- a/maven/src/test/scala/stryker4s/maven/runner/MavenMutantRunnerTest.scala
+++ b/maven/src/test/scala/stryker4s/maven/runner/MavenMutantRunnerTest.scala
@@ -1,6 +1,7 @@
 package stryker4s.maven.runner
 
 import better.files._
+import org.apache.maven.model.Profile
 import org.apache.maven.project.MavenProject
 import org.apache.maven.shared.invoker._
 import org.mockito.captor.ArgCaptor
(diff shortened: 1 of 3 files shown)


maven/src/test/scala/stryker4s/maven/runner/MavenMutantRunnerTest.scala
```diff
@@ -103,6 +104,25 @@ class MavenMutantRunnerTest extends Stryker4sSuite with MockitoSugar {
...
+
+    it("should propagate active profiles") {
+      val invokerMock = mock[Invoker]
+      val mockResult = mock[InvocationResult]
+      when(mockResult.getExitCode).thenReturn(0)
+      when(invokerMock.execute(any)).thenReturn(mockResult)
+      val captor = ArgCaptor[InvocationRequest]
+      val mavenProject = new MavenProject()
+      val profile = new Profile()
+      profile.setId("best-profile-ever")
```
hugo-vrijswijk, line 116: Best profile name ever!

jozic:
@hugo-vrijswijk 
changes looks good. 
I haven't tried activating profile via command line, but it sounds logical

PR #565, opened 2020-09-01
Filter out mutants that are in `val`s in an object

hugo-vrijswijk:
With #492, mutations that are initialized the moment the JVM is started cannot be tested. Examples of these mutations are when a `val` is directly inside an `object` like so:

```scala
object Foo {
  val bar = "42"
}
```

Because the JVM is only started once, the active mutation is checked and activated only once (on initialization), subsequent mutation runs will never properly activate the mutation and the mutant will always be `Survived`.

This PR filters out mutations when they are directly a `val` inside an `object`. This is a 'best guess' at filtering out these situations. If the mutation is inside a function it is not filtered out:

```scala
object Foo {
  // Filtered out
  val bar = "42"

  // Filtered out
  val bar2 = a.b("42")

  // Filtered out
  val bar3 = new Person("42")

  // Filtered out
  val bar4 = {
    val a = calculation()
    a + "42"
  }

  // Not filtered out
  val bar5 = () => "42"
}
```

@nicojs Thoughts on the solution (not necessarily the implementation)? I know you've run into something similar with mutation switching for Stryker JS.

### DIFF
--- a/core/src/main/scala/stryker4s/extension/TreeExtensions.scala
+++ b/core/src/main/scala/stryker4s/extension/TreeExtensions.scala
@@ -7,13 +7,16 @@ import scala.reflect.ClassTag
 import scala.util.Try
 
 object TreeExtensions {
-  @tailrec
-  private def mapParent[T <: Tree, U](tree: Tree, ifFound: T => U, notFound: => U)(implicit classTag: ClassTag[T]): U =
-    tree.parent match {
-      case Some(value: T)   => ifFound(value)
-      case Some(otherValue) => mapParent(otherValue, ifFound, notFound)
-      case _                => notFound
-    }
+
+  implicit class MapParentExtension(tree: Tree) {
+    @tailrec
+    final def mapParent[T <: Tree, U](ifFound: T => U, notFound: => U)(implicit classTag: ClassTag[T]): U =
+      tree.parent match {
+        case Some(value: T)   => ifFound(value)
+        case Some(otherValue) => otherValue.mapParent(ifFound, notFound)
+        case _                => notFound
+      }
+  }
 
   implicit class TopStatementExtension(thisTerm: Term) {
 
@@ -41,7 +44,7 @@ object TreeExtensions {
           .flatMap(findParent[Term])
 
       private def findParent[T <: Tree](tree: Tree)(implicit classTag: ClassTag[T]): Option[T] =
-        mapParent[T, Option[T]](tree, Some(_), None)
+        tree.mapParent[T, Option[T]](Some(_), None)
     }
 
     /** Extractor object to check if the direct parent of the [[scala.meta.Term]] is a 'full statement'
@@ -102,7 +105,7 @@ object TreeExtensions {
       * Recursively going up the tree until an annotation is found.
       */
     final def isIn[T <: Tree](implicit classTag: ClassTag[T]): Boolean =
-      mapParent[T, Boolean](thisTree, _ => true, false)
+      thisTree.mapParent[T, Boolean](_ => true, false)
   }
 
   implicit class PathToRoot(thisTree: Tree) {
--- a/core/src/main/scala/stryker4s/mutants/findmutants/MutantMatcher.scala
+++ b/core/src/main/scala/stryker4s/mutants/findmutants/MutantMatcher.scala
@@ -6,6 +6,7 @@ import stryker4s.config.Config
 import stryker4s.extension.TreeExtensions.{GetMods, PathToRoot, TreeIsInExtension}
 import stryker4s.extension.mutationtype._
 import stryker4s.model.Mutant
+import stryker4s.extension.TreeExtensions.MapParentExtension
 
 class MutantMatcher()(implicit config: Config) {
   private[this] val ids = Iterator.from(0)
@@ -91,6 +92,7 @@ class MutantMatcher()(implicit config: Config) {
             else
               Some(Mutant(ids.next(), original, mutationToTerm(mutated), mutated))
           }
+          .filterNot(_.exists(isObjectVal))
       }
 
     private def ifNotInAnnotation(maybeMutants: => Seq[Option[Mutant]]): Seq[Option[Mutant]] = {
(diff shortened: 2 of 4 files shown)


nicojs:
In your example `val bar5 = () => "42"` isn't filtered out. But consider this:

```scala
object Qux {
   qux = Foo.bar5()
}
```

Aha! It should have been filtered out after all!

There is no way to determine what are "static" mutants and which are non-static other than running the code. This is what Stryker does. If code is executed before tests are run, then those mutants are considered "static" and a reload of all files is done (in your case that would mean unload the files from the JVM, or start a new JVM).

@Mobrockers or @richardwerkman how does Stryker.NET deal with these "static" mutants? Maybe we can write a piece on it in the Stryker handbook and explain how each Stryker platform deals with it?

Mobrockers:
@dupdob can best answer that question

dupdob:
I am not sure the situation with Stryker.Net and Stryker4s are similar for statics, especially since I know very little of Scala.

Stryker.NET does not filter out 'static mutants', but they are identified and are systematically ran against all tests (in isolated test session). Otherwise, we would get false positive (fake survivors) because the static mutants would tested only against the first test needing them (in .Net, statics are lazy initiated, typically when the class is loaded).

So in .Net, static mutants are only an issue for coverage based optimisation, otherwise everything is fine.

They are detected with two mechanism:
- static analysis: when mutating the project, Stryker.Net marks mutants that are obvious static constructs
- runtime analysis: static constructs with method calls are instrumented, so when coverage analysis is run, we can mark any mutants that is called from a static constructs.


Note that in general, statics lead to a violation of the number one rule of TDD: having isolated tests.
So I strive to have tests for those static constructs where the static is forcibly reinitialised (in the setup phase).

nicojs:
Hi @dupdob 👋 thanks for pitching in.

> I am not sure the situation with Stryker.Net and Stryker4s are similar for statics, especially since I know very little of Scala.

Yes, they are. Both are compiled languages with static, dynamic code and static&dynamic code. 

> They are detected with two mechanisms:
> * static analysis: [...]
> * runtime analysis: [...]


What @hugo-vrijswijk uses here is the static analysis approach. It needs runtime analysis to make sure (see my example with `Qux`). It should also should not simply filter out those mutants IMHO. It should _at leat_ report them as `ignored`. Preferably you would still be able to test them but with a performance penalty. 

In general, maybe we should allow for a `--ignore-static-mutants` flag (or another name) in all versions of Stryker. That way you can choose to not test static mutants (instead reporting them as ignored).

> There is an issue for that and I plan to work on an improved roll back logic in the coming weeks

Sounds awesome! GL

> Note that in general, statics lead to a violation of the number one rule of TDD: having isolated tests.

True. But we shouldn't force people to write all their code with this best practice in mind. I'm sure you agree 🙏

dupdob:
@nicojs : 
I agree 😉. Here is what I have in mind:
When I test properly my static code, I don't need Stryker to extensively test the static mutants, so I would rather have: `--trust-coverage-for-static-mutants` than '--ignore....'

But this is a detail, and I agree with your first comment: this is exactly what Stryker-Mutator.Net does for static.

hugo-vrijswijk:
Thanks for the input, everyone! Great to see some cross-Stryker discussions 😀.

As far as reporting goes I am not sure 'static' mutants should be reported as ignored. We've defined 'ignored' as: "The mutant was not tested because the config of the user asked for it to be ignored". This is different because the user did not ask for static mutants to be skipped, and the reason for skipping it is purely technical. Stryker4s supports 'static' mutants in the current version, but not in the next. At most I think a warning should be logged.

@nicojs is correct in that this in Scala:

```scala
object Foo {
  val bar = 5
}
```

Is equivalent to the C#:

```csharp
static class Foo
{
    public static int bar = 5;
}
```

Scala seems to only initialize the value when it is first called, not when the JVM is started.

The snippet from Nicojs could be considered 'static' because `qux` is filled with a static value. However, if `Foo.bar5()` is not static it can still be tested separately as a non-static value. In this case I would consider `qux` the static mutant, not `Foo.bar5()`.

> @nicojs: If mutated code is executed before tests are run, then those mutants are considered "static" and a reload of all files is done (in your case that would mean unload the files from the JVM, or start a new JVM).

> @dupdob: static constructs with method calls are instrumented, so when coverage analysis is run, we can mark any mutants that is called from a static constructs.

The issue here is that values can be lazily-initiated. So any coverage analysis wouldn't be called at all until tests are run. Another downside is that restarting the JVM or reloading all files is very expensive.

A downside of static code (AST) analysis is also that if something 'looks' like static it doesn't have to be. Consider this Scala example:

```scala
object Foo {
  val five = MyClass(5)
  val six = IO(6)
}
```

Here `five` is a static mutant that would be easy to statically detect. But `six` 'suspends' the value 6 in something called an IO. To make a long story short, the `6` will be re-evaluated again every time `six` is called (like a method), so it is perfectly fine to mutate this code. Static analysis (like this PR) would wrongfully filter out `six`. This might be Scala-specific, but it I can imagine it is possible to do similar things in other languages (properties in C#, maybe?) and something Stryker would have to deal with.

I agree that the only really reliable way to find 'static' mutants is with runtime analysis. I've been thinking about it some more over the past couple days, and I think the best way is by doing an initial test run twice using coverage analysis:

When doing coverage analysis twice with two initial test-runs you can look at the difference in what mutants are covered by the two test-runs. All mutants that were covered in the first testrun, but not in the second _have_ to be static. Those can then be marked as such. It's important that this is done with no mutation active as you cannot know if it will cause a test-failure in advance, and a mutation can cause a difference in code-path and thus different mutants called.

The code from my example can then be mutated to: (pseudocode)

```scala
object Foo {
  val bar = stryker4s.coverMutant(id = "1").getActiveMutant() match {
    case "1" => ""
    case _ => "42"
  }
}
```

The testrunner would do something like: (pseudocode)

```scala
val initialTestRunCoveredMutants = runInitialTestRun()
resetMutantCoverage()
val secondInitialTestRunCoveredMutants = runInitialTestRun()
val staticMutants = second.diff(first) // All called mutants that are in the second list but not in the first
```

From what I can see this approach would cover (no pun intended) all scenarios of detecting static mutants with the only downside being some more complex coverage instrumentation needed and having to do the initial test run twice (with hundreds of test runs that shouldn't be too big of an impact).

hugo-vrijswijk:
Closing this PR for now as we're going in a different direction

PR #622, opened 2020-10-31
HTML report body changes with theme

wijtserekker:
#### What it does
When you change the theme of the HTML report, the background of the parent HTML page changes also.

#### How it works
The HTML report component dispatches a theme-changed event when the theme has changed. This solution listens for that event and changes the background of the body accordingly.


hugo-vrijswijk:
Awesome, thanks!

PR #908, opened 2021-07-15
Simplest way make config file selectable

cbrunnkvist:
#### What it does

Makes the config file path less hardcoded, e.g. allows `sbt -Dstryker4s.config.file=mod-x/stryker4s.conf stryker`. Not sure if we should go the "make it a parameter of the sbt task"-route, but _at least_ this is a much less invasive fix, and is already much better than having to move config files around in the filesystem, just in order to switch between subproject runs.

#### How it works

Not sure if it works - this PR comes straight from the GitHub file editor ;)

### DIFF
--- a/core/src/main/scala/stryker4s/config/ConfigReader.scala
+++ b/core/src/main/scala/stryker4s/config/ConfigReader.scala
@@ -16,7 +16,9 @@ object ConfigReader {
 
   implicit val hint: ProductHint[Config] = ProductHint[Config](allowUnknownKeys = false)
 
-  def readConfig(confSource: ConfigSource = ConfigSource.file("stryker4s.conf"))(implicit log: Logger): Config = {
+  lazy val defaultConfig = ConfigSource.file(System.getProperty("stryker4s.config.file", "stryker4s.conf"))
+
+  def readConfig(confSource: ConfigSource = defaultConfig)(implicit log: Logger): Config = {
     Reader
       .withoutRecovery[Config](confSource)
       .recoverWithReader(Failure.onUnknownKey)
@@ -25,7 +27,7 @@ object ConfigReader {
   }
 
   def readConfigOfType[T](
-      confSource: ConfigSource = ConfigSource.file("stryker4s.conf")
+      confSource: ConfigSource = defaultConfig
   )(implicit log: Logger, pureconfig: PureConfigReader[T]): Either[ConfigReaderFailures, T] =
     Reader.withoutRecovery[T](confSource).tryRead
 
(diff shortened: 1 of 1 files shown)


core/src/main/scala/stryker4s/config/ConfigReader.scala
```diff
@@ -75,7 +77,7 @@ object ConfigReader {
     /** Attempt to read a config
       */
     def tryRead: Reader.Result[T] = {
-      log.info(s"Attempting to read config from stryker4s.conf")
+      log.info(s"Attempting to read config from ${defaultConfig}")
```
cbrunnkvist, line 80: This will not toString nicely... Needs to separate _file name_ from _config object_.

hugo-vrijswijk: Can you put the `getProperty` or default in a separate variable to properly log the file location?

hugo-vrijswijk (commented):
Awesome! I think a system property is a good place to put this functionality. Could you add some tests and maybe a couple of lines in the configuration docs? With that, this is ready to merge, I think!

cbrunnkvist:
Ok, I'll have a look. Btw what's your typical workflow for testing the plugin locally - do you go through `publishLocal` and then a `clean update` in an example project, or is there some shortcut to way to achieve the same?

cbrunnkvist:
Please help me refresh my memory, what will be the difference in behavior, if I have set base-dir to `module-a` and:

1. I run `stryker` (=in the root project)
2. or I run `module-a/stryker`?

hugo-vrijswijk:
> Btw what's your typical workflow for testing the plugin locally

Yes, `publishLocal`, then change the version number in `plugins.sbt` and run Stryker. No need to clean anything.

> Please help me refresh my memory, what will be the difference in behavior, if I have set base-dir to module-a and

On multi-module projects, running `stryker` in the 'root' sbt project will try to run Stryker4s against all projects simultaneously, which... doesn't work. The best way is `project module-a; stryker`. That has different behaviour from `module-a/stryker` (not sure what exactly the difference is, though)

cbrunnkvist:
Note to self: also update the ./README.md

hugo-vrijswijk:
@cbrunnkvist any update on this?

cbrunnkvist:
I could never get it to compile once I had made some change. Most likely something underlated e.g. caused by bloop / metals whatever. I'm going to give it a sec...third shot.

hugo-vrijswijk:
The configuration reading has changed a lot since. You can now configure via sbt on a per-project basis as well. If this is still needed, let me know

PR #912, opened 2021-07-21
Use loopback address for testrunner comms.

### LINKED ISSUE #911: Timeout connecting to testrunner
The testrunner sub-processes seem to be binding to a wildcard address, which is both potentially a security issue plus seems to interfere with different networking situations such as when connected through a VPN etc.

Ref. https://gitter.im/stryker-mutator/stryker4s?at=60dd5bb6368d0e1d9a36f3a8

It should be enough (& equally portable) to bind to localhost/loopback address instead.

cbrunnkvist:
(fixes #911)


hugo-vrijswijk:
Thank you for the find and the quick fix!

hugo-vrijswijk (approved):

PR #979, opened 2021-10-01
Two small documentation fixes

MartinWelgemoed:

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ You can then run Stryker4s with the command `mvn stryker4s:run`. Note that this 
 
 ## Pre-release versions
 
-We also publish SNAPSHOT versions of each commit on master. To use a pre-release, add the following setting to your `plugins.sbt`:
+We also publish SNAPSHOT versions of each commit on master. To use a pre-release, add the following setting to your `build.sbt`:
 
 ```scala
 resolvers += Resolver.sonatypeRepo("snapshots")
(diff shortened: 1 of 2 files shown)


sbt/src/sbt-test/sbt-stryker4s/test-1/README.md
```diff
@@ -2,6 +2,8 @@
 
 This is a test project for the Stryker4s SBT plugin.
 
-To run it, first run `sbt sbt-stryker4s2_12/publishLocal` in the root Stryker4s project.
+To run it, first
+run `sbt stryker4s-core2_12/publishLocal; stryker4s-api2_12/publishLocal; sbt-stryker4s-testrunner2_12/publishLocal` in
```
MartinWelgemoed, line 6: Following the readme didn't work locally, so I peeked at what the CI did and copied that, which worked.

hugo-vrijswijk: I think this can also be replaed with the command `sbt publishPluginLocal` which does all that (I think at least for 2.13 projects, don't know if it works for 2.12 projects)

README.md
```diff
@@ -55,7 +55,7 @@ You can then run Stryker4s with the command `mvn stryker4s:run`. Note that this
 
 ## Pre-release versions
 
-We also publish SNAPSHOT versions of each commit on master. To use a pre-release, add the following setting to your `plugins.sbt`:
+We also publish SNAPSHOT versions of each commit on master. To use a pre-release, add the following setting to your `build.sbt`:
```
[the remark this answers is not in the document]
hugo-vrijswijk: Shouldn't it be in both versions? The reason it's needed in `build.sbt` is the test-runner dependency that's added. But I think it'd also be possible to recognize if a snapshot is running and add the snapshot resolver to the build scope

PR #1262, opened 2022-12-01
sbt: enable local publish for stryker4s-command-runner

gergelytraveltime:
This is useful when running local tests for stryker4s-command-runner.
Make publishPluginLocal also publish stryker4s-command-runner for Scala 2.12 and 2.13.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -15,6 +15,11 @@ lazy val root = (project withId "stryker4s" in file("."))
     addCommandAlias(
       "publishM2Local",
       "set ThisBuild / version := \"SET-BY-SBT-SNAPSHOT\"; stryker4s-core/publishM2; stryker4s-api/publishM2"
+    ),
+    // Publish to .ivy folder for command runner local testing
+    addCommandAlias(
+      "publishCommandRunnerLocal",
+      "set ThisBuild / version := \"0.0.0-TEST-SNAPSHOT\"; stryker4s-core/publishLocal; stryker4s-api/publishLocal; stryker4s-command-runner/publishLocal"
     )
   )
   .aggregate(


build.sbt
```diff
@@ -9,7 +9,7 @@ lazy val root = (project withId "stryker4s" in file("."))
     // Publish locally for sbt plugin testing
     addCommandAlias(
       "publishPluginLocal",
-      "set ThisBuild / version := \"0.0.0-TEST-SNAPSHOT\"; stryker4s-core2_12/publishLocal; stryker4s-api2_12/publishLocal; sbt-stryker4s2_12/publishLocal; stryker4s-api/publishLocal; sbt-stryker4s-testrunner/publishLocal"
+      "set ThisBuild / version := \"0.0.0-TEST-SNAPSHOT\"; stryker4s-core2_12/publishLocal; stryker4s-api2_12/publishLocal; sbt-stryker4s2_12/publishLocal; stryker4s-api/publishLocal; sbt-stryker4s-testrunner/publishLocal; stryker4s-command-runner/publishLocal; stryker4s-command-runner2_12/publishLocal"
```
hugo-vrijswijk, line 12: I think this should be a separate command. There is already a separate one for Maven, so it would make sense to have one for the command-runner too. Then only publishing 2.13 versions should also be enough (it doesn't matter what your own Scala version is for the command-runner, so you don't need 2.12):

```
stryker4s-core/publishLocal; stryker4s-api/publishLocal; stryker4s-command-runner/publishLocal
```

gergelytraveltime: Done.

build.sbt
```diff
@@ -15,6 +15,11 @@ lazy val root = (project withId "stryker4s" in file("."))
     addCommandAlias(
       "publishM2Local",
       "set ThisBuild / version := \"SET-BY-SBT-SNAPSHOT\"; stryker4s-core/publishM2; stryker4s-api/publishM2"
+    ),
+    // Publish to .ivy folder for local testing
+    addCommandAlias(
+      "publishLocal",
```
hugo-vrijswijk, line 21:
```suggestion
      "publishCommandRunnerLocal",
```

publishLocal is an existing sbt task, so I'd like to keep the distinction of what this publishes

gergelytraveltime: Thanks. I think I got it finally. We need a special publish command for command-runner, but it also needs to publish core and api as those are dependencies.

gergelytraveltime:
Updated with requested changes.

hugo-vrijswijk:
Thanks! Very useful

Issue #668, opened 2020-12-17
Config option to not delete temp dir after running


hugo-vrijswijk:
Right now the temp dir is deleted after running using a FS2 tempDir [`Resource`](https://typelevel.org/cats-effect/docs/std/resource). This is done in the [MutantRunner](https://github.com/stryker-mutator/stryker4s/blob/master/core/src/main/scala/stryker4s/run/MutantRunner.scala#L53). 
It'd be nice to have a config option to keep the generated temp dir after running for debugging purposes to look at the generated code. An option similar to [Strykers](https://stryker-mutator.io/docs/stryker/configuration#cleantempdir-boolean) would be preferred.

Instead of `Files[IO].tempDirectory()` which returns a resource that removes the directory when closed, we can use `Files[IO].createTempDirectory` when the config option is enabled which only creates, but not removes

cbrunnkvist:
Yeah that would probably be useful, not in normal use but debugging. Perhaps a `stryker4s.skip-cleanup` or `stryker4s.debug.skip-cleanup` config boolean..?

hugo-vrijswijk:
I think it's good to keep in line with Stryker-js, which uses [`cleanTempDir`](https://stryker-mutator.io/docs/stryker-js/configuration/#cleantempdir-boolean), so the Stryker4s option would be `clean-temp-dir`

hugo-vrijswijk:
Fixed in #1263

PR #1263, opened 2022-12-01
staticTmpDir: enable forcing tmpDir to a static path

gergelytraveltime:
Added staticTmpDir option to Config.
This will force tmpDir to be created at a static path. Tmp dir will be autodeleted at exit, maybe that
`staticTmpDirLeaveAfterError` was defined, then we'll show a warning that it needs to be manually cleaned up.

This helps with using stryker4s with Bazel, as Bazel uses the path to the workspace (folder) to determine where to store the local cache (hashes the path to create the sandbox's name).
Enabled not autoremoving the folder to be able to debug issues (as Bazel support is early-stage).

### DIFF
--- a/core/src/main/scala/stryker4s/config/Config.scala
+++ b/core/src/main/scala/stryker4s/config/Config.scala
@@ -21,7 +21,9 @@ final case class Config(
     legacyTestRunner: Boolean = false,
     scalaDialect: Dialect = dialects.Scala3,
     concurrency: Int = Config.defaultConcurrency,
-    debug: DebugOptions = DebugOptions()
+    debug: DebugOptions = DebugOptions(),
+    staticTmpDir: Boolean = false,
+    staticTmpDirLeaveAfterError: Boolean = false
 )
 
 object Config extends pure.ConfigConfigReader with circe.ConfigEncoder {
--- a/core/src/main/scala/stryker4s/run/MutantRunner.scala
+++ b/core/src/main/scala/stryker4s/run/MutantRunner.scala
@@ -69,9 +69,36 @@ class MutantRunner(
     val targetDir = config.baseDir / "target"
     for {
       _ <- Resource.eval(Files[IO].createDirectories(targetDir))
-      tmpDir <- Files[IO].tempDirectory(Some(targetDir), "stryker4s-", None)
+      tmpDir <- prepareTmpDir(targetDir)
       _ <- Resource.eval(setupFiles(tmpDir, mutatedFiles.toSeq))
     } yield tmpDir
+  }
+
+  private def prepareTmpDir(targetDir: Path): Resource[IO, Path] = {
+    if (config.staticTmpDir) {
+      val staticTmpDir = targetDir / "stryker4s-tmpDir"
+      Resource
+        .eval(Files[IO].createDirectory(staticTmpDir).map(_ => staticTmpDir))
+        .onFinalizeCase(staticTmpDirFinalizeCase(staticTmpDir))
+    } else {
+      Files[IO].tempDirectory(Some(targetDir), "stryker4s-", None)
+    }
+  }
+
+  private def staticTmpDirFinalizeCase(staticTmpDir: Path): Resource.ExitCase => IO[Unit] = {
+    case Resource.ExitCase.Succeeded | Resource.ExitCase.Canceled =>
+      Files[IO].deleteRecursively(staticTmpDir)
+    case _: Resource.ExitCase.Errored =>
+      // Enable the user do some manual actions on the staticTmpDir before she retries.
+      if (config.staticTmpDirLeaveAfterError) {
+        IO(
+          log.warn(
+            s"Not deleting $staticTmpDir after error (turn off staticTmpDirLeaveAfterError to disable this). Please clean it up manually."
+          )
+        )
+      } else {
+        Files[IO].deleteRecursively(staticTmpDir)
+      }
   }
 
   private def setupFiles(tmpDir: Path, mutatedFiles: Seq[MutatedFile]): IO[Unit] =


hugo-vrijswijk:
Thanks for the PR! This would also close #668 I believe. Could you name the option the same as in StrykerJS? So `clean-temp-dir` with default false (except on error is a nice touch 👌)? Then it could also be enabled for the 'normal' temp directory setup.

core/src/main/scala/stryker4s/run/MutantRunner.scala
```diff
@@ -69,11 +69,38 @@ class MutantRunner(
...
       _ <- Resource.eval(setupFiles(tmpDir, mutatedFiles.toSeq))
     } yield tmpDir
   }
 
+  private def prepareTmpDir(targetDir: Path): Resource[IO, Path] = {
+    if (config.staticTmpDir) {
+      val staticTmpDir = targetDir / "stryker4s-tmpDir"
+      Resource
+        .eval(Files[IO].createDirectory(staticTmpDir).map(_ => staticTmpDir))
+        .onFinalizeCase(staticTmpDirFinalizeCase(staticTmpDir))
```
hugo-vrijswijk, line 82: I think what you want here is `Resource.makeCase` with the first argument being creating the directory, and the second handling deleting it when the resource is done

hugo-vrijswijk (changes requested):

hugo-vrijswijk:
> > Thanks for the PR! This would also close #668 I believe. Could you name the option the same as in StrykerJS? So `clean-temp-dir` with default false (except on error is a nice touch ok_hand)? Then it could also be enabled for the 'normal' temp directory setup.
> 
> I guess `clean-temp-dir` with default `true`, right?

Yes! Sorry, I didn't sleep much last night haha

core/src/main/scala/stryker4s/run/MutantRunner.scala
```diff
@@ -69,11 +69,38 @@ class MutantRunner(
...
+      Files[IO].tempDirectory(Some(targetDir), "stryker4s-", None)
+    }
+  }
+
+  private def staticTmpDirFinalizeCase(staticTmpDir: Path): Resource.ExitCase => IO[Unit] = {
+    case Resource.ExitCase.Succeeded | Resource.ExitCase.Canceled =>
+      Files[IO].deleteRecursively(staticTmpDir)
+    case _: Resource.ExitCase.Errored =>
+      // Enable the user do some manual actions on the staticTmpDir before she retries.
+      if (config.staticTmpDirLeaveAfterError) {
```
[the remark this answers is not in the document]
hugo-vrijswijk: `Files[IO].createTempDirectory` should do the trick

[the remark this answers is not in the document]
gergelytraveltime: Ah, I found it, `createTempDirectory`

core/src/main/scala/stryker4s/run/MutantRunner.scala
gergelytraveltime: Done.

gergelytraveltime:
> Added staticTmpDir option to Config. This will force tmpDir to be created at a static path. Tmp dir will be autodeleted at exit, maybe that `staticTmpDirLeaveAfterError` was defined, then we'll show a warning that it needs to be manually cleaned up.
> 
> This helps with using stryker4s with Bazel, as Bazel uses the path to the workspace (folder) to determine where to store the local cache (hashes the path to create the sandbox's name). Enabled not autoremoving the folder to be able to debug issues (as Bazel support is early-stage).

Updated description:

    staticTmpDir: enable forcing tmpDir to a static path
    
    Added staticTmpDir option to Config.
    This will force tmpDir to be created at a static path.
    
    This helps with using stryker4s with Bazel, as Bazel uses the path to the
    workspace (folder) to determine where to store the local cache (hashes
    the path to create the sandbox's name).
    
    -----
    
    Enabled not autoremoving the folder to be able to debug issues:
    
    Tmp dir will be autodeleted at exit, maybe that
    `clearTmpDir` was disabled, then we'll show a warning that
    it needs to be manually cleaned up.
    On error the tmpDir is not deleted (whether or not static).

core/src/main/scala/stryker4s/run/MutantRunner.scala
```diff
@@ -69,11 +69,41 @@ class MutantRunner(
...
-      tmpDir <- Files[IO].tempDirectory(Some(targetDir), "stryker4s-", None)
+      tmpDir <- prepareTmpDir(targetDir)
       _ <- Resource.eval(setupFiles(tmpDir, mutatedFiles.toSeq))
     } yield tmpDir
   }
 
+  private def prepareTmpDir(targetDir: Path): Resource[IO, Path] = {
+    val tmpDirCreated = if (config.staticTmpDir) {
+      val staticTmpDir = targetDir / "stryker4s-tmpDir"
+      Files[IO].createDirectory(staticTmpDir).map(_ => staticTmpDir)
```
hugo-vrijswijk, line 80:
```suggestion
      Files[IO].createDirectory(staticTmpDir).as(staticTmpDir)
```

core/src/main/scala/stryker4s/config/Config.scala
```diff
@@ -21,7 +21,9 @@ final case class Config(
     legacyTestRunner: Boolean = false,
     scalaDialect: Dialect = dialects.Scala3,
     concurrency: Int = Config.defaultConcurrency,
-    debug: DebugOptions = DebugOptions()
+    debug: DebugOptions = DebugOptions(),
+    staticTmpDir: Boolean = false,
```
hugo-vrijswijk, line 25: Could you also add a test for the new options in `ConfigReaderTest` and add docs for them in `docs/configuration.md`? Then I think we can merge this!

hugo-vrijswijk (commented):
Looks good!! One minor style thing, and a question about tests and docs for the new config options, but then we can merge this

core/src/main/scala/stryker4s/run/MutantRunner.scala
```diff
@@ -69,11 +69,41 @@ class MutantRunner(
...
+    } else {
+      Files[IO].createTempDirectory(Some(targetDir), "stryker4s-", None)
+    }
+    Resource.makeCase(tmpDirCreated)(tmpDirFinalizeCase)
+  }
+
+  private val tmpDirFinalizeCase: (Path, Resource.ExitCase) => IO[Unit] = {
```
hugo-vrijswijk, line 87: Is there a specific reason you made this a `val` instead of a `def`? I think it would be slightly cleaner as a def since you wouldn't have to have the `case`

gergelytraveltime: I think I still need to have the case.

hugo-vrijswijk: Yes, but couldn't it be `def tmpDirFinalizeCase(exitCase....): IO[Unit]`?

core/src/main/scala/stryker4s/config/Config.scala
gergelytraveltime: Done.

core/src/main/scala/stryker4s/run/MutantRunner.scala
gergelytraveltime: Switched to a def in any case.

gergelytraveltime: > I think I still need to have the case.

I think what I don't need to have is a `match`, but it's also not necessary when defining as a `val`.

docs/configuration.md
```diff
@@ -120,6 +120,28 @@ Set the Scala dialect that should be used for parsing Scala files. The default i
 
 Valid values are Scala-versions without a patch version (`scala2.12`, `212`, `2.12`, `2`). If you use `-Xsource:3` you can use `scala212source3` or `scala213source3`. The full list can be found [here](https://github.com/stryker-mutator/stryker4s/blob/master/core/src/main/scala/stryker4s/config/pure/ConfigConfigReader.scala#L80-L84).
 
+### `static-tmp-dir` [`boolean`]
+
+**Config file:** `static-tmp-dir: true`
+**Default value:** `false`
```
hugo-vrijswijk, line 126:
```suggestion
**Default value:** `false`
**Since:** `v0.15.0`  
```

docs/configuration.md
```diff
@@ -120,6 +120,28 @@ Set the Scala dialect that should be used for parsing Scala files. The default i
...
+
+This will force the temporary dir to be created at a static path (`target/stryker4s-tmpDir`).
+
+Recommended when using stryker4s with Bazel (Bazel uses the path to the workspace to determine where to store the
+local cache - hashes the path to create the sandbox's name) as it speeds up mutation testing greatly.
+
+### `clean-tmp-dir` [`boolean`]
+
+**Config file:** `clean-tmp-dir: false`
+**Default value:** `true`
```
hugo-vrijswijk, line 137:
```suggestion
**Default value:** `true`
**Since:** `v0.15.0`  
```

core/src/main/scala/stryker4s/run/MutantRunner.scala
hugo-vrijswijk: Like I said, I didn't sleep much last night, but i see now why the `case` is there

docs/configuration.md
gergelytraveltime: Ok.

docs/configuration.md
gergelytraveltime: Ok.

gergelytraveltime:
Applied requested docs change and added new tests for `MutantRunner`.

core/src/test/scala/stryker4s/run/MutantRunnerTest.scala
```diff
@@ -95,5 +107,104 @@ class MutantRunnerTest extends Stryker4sIOSuite with MockitoIOSuite with LogMatc
...
+      val file = FileUtil.getResource("scalaFiles/simpleFile.scala")
+      val mutants = Seq(mutant)
+      val mutatedFile = MutatedFile(file, q"def foo = 4", mutants, Seq.empty, 0)
+
+      sut(_ => IO.pure(List(mutatedFile))).asserting { result =>
+        val captor = ArgCaptor[FinishedRunEvent]
+        verify(reporterMock, times(1)).onRunFinished(captor.capture)
+        val runReport = captor.value.report.files.loneElement
+
+        "Setting up mutated environment..." shouldBe loggedAsInfo
```
gergelytraveltime, line 134: Is adding such simple assertions here ok? Or test for even less, as mutation are checked in another test case?

gergelytraveltime:
Anything else necessary @hugo-vrijswijk?

core/src/test/scala/stryker4s/run/MutantRunnerTest.scala
hugo-vrijswijk: To keep the test clean, it's best to only test for the changes that this test has compared to the others. So since logging is already covered in other tests you don't need to assert for it again

gergelytraveltime: Removed some assertions. Should I remove even more?

hugo-vrijswijk (approved):
Thanks so much for this PR and the clean solution!

gergelytraveltime:
> Thanks so much for this PR and the clean solution! I'll merge it now, and if you really can't wait, you can use a snapshot release soon. Otherwise this will be in the next release (which shouldn't be too far away)

Thanks for all your help and assistance!

PR #1291, opened 2023-02-03
Converting Stryker4S into Stryker4JVM supporting Scala and Kotlin

mmulder135:
Stryker4S provides mutation testing for Scala, while Stryker4K does this for Kotlin. Since these projects are developed in parallel, and they are both JVM languages, they have large parts of code with exactly the same functionality (such as file IO, logging and reporting). The main problem to be solved is thus the code duplication between the different code bases. In addition to this, we should try to ensure that the shared interface can also be implemented by other JVM languages at a later time.


README.md
```diff
@@ -16,63 +16,75 @@ For an introduction to mutation testing and Stryker's features, see [stryker-mut
...
-addSbtPlugin("io.stryker-mutator" % "sbt-stryker4s" % stryker4sVersion)
+addSbtPlugin("io.stryker-mutator" % "stryker4jvm-plugin-sbt" % stryker4jvmVersion)
 ```
 
-After adding the plugin, Stryker4s can be used by running `sbt stryker` in the root of your project.
+After adding the plugin, Stryker4jvm can be used by running `sbt stryker` in the root of your project.
 
 ### Multi-module projects
 
-Multi-module projects are not yet fully supported. However, there is a workaround you can use while we work on a better solution. Set the base-directory to the correct directory of the submodule with the [`base-dir` configuration setting](https://stryker-mutator.io/docs/stryker4s/configuration#base-dir-string). Then you can run `sbt "project yourSubmoduleNameHere" stryker` to set the active project and run Stryker4s.
```
hugo-vrijswijk, line 35: In general links should still direct people to the website. Maybe once we've finalized on the name we can also update the website directory (and redirect the old links)

docs/configuration.md
```diff
@@ -118,7 +118,7 @@ Settings for the dashboard [reporter](#reporters-string). See the [dashboard doc
 
 Set the Scala dialect that should be used for parsing Scala files. The default is Scala 2.13 with `-XSource:3` as this has the widest compatibility. If you are running into issues with parsing older unsupported Scala syntax that we forgot about you can change this value.
 
-Valid values are Scala-versions without a patch version (`scala2.12`, `212`, `2.12`, `2`, `3`, `3.2`). If you use `-Xsource:3` you can use `scala212source3` or `scala213source3`. The full list can be found [here](https://github.com/stryker-mutator/stryker4s/blob/master/core/src/main/scala/stryker4s/config/pure/ConfigConfigReader.scala#L74-L84).
+Valid values are Scala-versions without a patch version (`scala2.12`, `212`, `2.12`, `2`, `3`, `3.2`). If you use `-Xsource:3` you can use `scala212source3` or `scala213source3`. The full list can be found [here](./stryker4jvm-mutator-scala/src/main/scala/stryker4jvm/mutator/scala/ScalaMutatorProvider.scala).
```
hugo-vrijswijk, line 121: This link should still be an absolute link to github because a relative link doesn't work when just this file is hosted on the website. It's fine to either have the link point to your branch or have a broken link that will work when this is merged

docs/contributing.md
```diff
@@ -4,9 +4,38 @@ title: Contribute to Stryker4s
...
-
+## Developing on stryker4jvm:
+Currently, stryker4jvm is using 3 build tools for all its modules. Stryker4jvm-core is managed by maven,
+stryker4jvm-mutator-kotlin is managed by gradle and all other modules are managed by sbt. In order to build/test locally
+you need to take the following steps:
+
+Build has only been confirmed to work with:
+* Jvm 15.0.2,
+* Maven 3.6.3,
+* Kotlin 1.4.20,
```
hugo-vrijswijk, line 15: The current latest release of Kotlin is 1.8.10. I'd expect Stryker4jvm to always be tested against the latest (or LTS) version. So, this would really have to be updated before this can be merged (same with the other versions). Scala also isn't mentioned here, but I would suggest not keeping this list, as it will probably rapidly become out of date

project/Dependencies.scala
```diff
@@ -10,10 +10,6 @@ object Dependencies {
       */
     val crossScalaVersions = Seq(scala213, scala212)
 
-    /** Fuller cross-versions (used for injected packages like stryker4s-api and sbt-stryker4s-testrunner)
-      */
-    val fullCrossScalaVersions = crossScalaVersions ++ Seq(scala3)
```
hugo-vrijswijk, line 15: Why was this removed? The injected sbt testrunner still needs to be published for the fullest range possible (so that includes Scala 3)

project/Release.scala
```diff
@@ -19,12 +16,12 @@ object Release {
 
   lazy val releaseCommands: Setting[Seq[Command]] = commands ++= Seq(
     // Called by sbt-ci-release
-    Command.command(stryker4sPublishSigned)(
```
hugo-vrijswijk, line 22: I think the release process is broken by this PR. Maybe an interim solution can be to add a todo so we know it has to be fixed after merging this PR?

stryker4jvm-core/merge-notes-java-conversion.txt
```diff
@@ -0,0 +1,35 @@
...
+Stryker4jvmException has been converted a java exception
+
+files:
+FileIO has been converted to a java interface;
+The class DisFileIO is reimplemented with pure java, in case that we still want IO[Unit] for fileIO in stryker4jvm
+we could implement a wrapper (should be possible).
+FileResolver has been ignored as it relies on fs2, this may instead be added to stryker4jvm
+
+Reporting:
+Converted scala Reporter to abstract java class without any IO (wrappers may be necessary in stryker4jvm if we wish to work with IO there);
```
hugo-vrijswijk, line 28: Could the reporters not also be inside stryker4jvm? They shouldn't be used by individual language mutators

sbt/src/main/scala/stryker4s/log/SbtLogger.scala
```diff
@@ -1,25 +0,0 @@
-package stryker4s.log
-
-import sbt.{Level as SbtLevel, Logger as SbtInternalLogger}
-
-class SbtLogger(sbtLogger: SbtInternalLogger) extends Logger {
-
-  def log(level: Level, msg: => String): Unit = sbtLogger.log(toSbtLevel(level), msg)
```
hugo-vrijswijk, line 7: I don't see this class moved anywhere. How does logging now work inside the plugins?

project/plugins.sbt
```diff
@@ -5,4 +5,7 @@ addSbtPlugin("com.eed3si9n" % "sbt-projectmatrix" % "0.9.0")
 
 // Protobuf plugin and its dependencies
 addSbtPlugin("com.thesamet" % "sbt-protoc" % "1.0.6")
-libraryDependencies += "com.thesamet.scalapb" %% "compilerplugin" % "0.11.12"
+libraryDependencies += "com.thesamet.scalapb" %% "compilerplugin" % "0.11.11"
+
+// Coverage
+addSbtPlugin("org.scoverage" % "sbt-scoverage" % "2.0.6")
```
hugo-vrijswijk, line 11: I don't think adding code coverage should really be a goal of this PR. It should probably be separate (or better, use mutation coverage)

stryker4jvm/src/main/scala/stryker4jvm/logging/FansiLogger.scala
```diff
@@ -0,0 +1,33 @@
+package stryker4jvm.logging
+import fansi.Str
+import stryker4jvm.core.logging.{LogLevel, Logger}
+
+class FansiLogger(val coreLogger: Logger) {
```
hugo-vrijswijk, line 5: Maybe a clean solution for the loggers could be to have the same Logger class in core with just `String`s in core, but to have this class (maybe renamed to ColorLogger) extend the base Logger and add the `Fansi` methods and color checking logic. Then the language mutators could have a `Logger` interface passed, while all the projects 'below' that would get a `ColorLogger` injected, which would be the same logger instantation, but just with a 'wider' interface

docs/contributing.md
```diff
@@ -4,9 +4,38 @@ title: Contribute to Stryker4s
 custom_edit_url: https://github.com/stryker-mutator/stryker4s/edit/master/docs/contributing.md
 sidebar_position: 3
 ---
-
-This is the contribution guide for Stryker4s. Great to have you here! Here are a few ways you can help to make this project better.
-
+## Developing on stryker4jvm:
+Currently, stryker4jvm is using 3 build tools for all its modules. Stryker4jvm-core is managed by maven,
```
hugo-vrijswijk, line 8: I think it would also help to have a picture (or ascii diagram) of the project layout as it's quite a bit more complex now

hugo-vrijswijk (commented):
Some remarks about the overall layout and some remarks from an initial glance. I haven't really looked into the code yet, I'll probably be able to do that in the coming days

sbt/src/main/scala/stryker4s/log/SbtLogger.scala
s2176416: SbtLogger is moved to module stryker4jvm-plugin-sbt. The SbtLogger class is private but a new FansiSbtLogger class has been created.

project/Dependencies.scala
s2176416: Is there a use-case for using fullCrossScalaVersions for api and testrunner? The sbt plugin which uses testrunner is only built with scala 2_12 and all other modules are built with 2_12 or 2_13.

hugo-vrijswijk: Yes, api and testrunner are injected libraries when starting a testrunner. It runs alongside the tests. So they need to be built on whatever scala version the end-users project is, otherwise they'll get conflicts of mismatched Scala versions

stryker4jvm-core/merge-notes-java-conversion.txt
s2176416: The merge-notes are outdated and are merely used during development of stryker4jvm to clarify where some things have been moved. They should not have ended up in the PR and will be removed.

stryker4jvm/src/main/scala/stryker4jvm/logging/FansiLogger.scala
s2176416: Because fansi.str automatically converts to strings it was unclear at times whether someone was actually using a 'FansiLogger' or a 'core Logger', for this reason it was decided to split it but the proposed solution should work as well.

.github/workflows/ci.yml
```diff
@@ -4,37 +4,60 @@ on: [push, pull_request]
...
         with:
-          distribution: 'temurin'
+          distribution: 'corretto'
           java-version: ${{ matrix.java-version }}
           cache: 'sbt'
-      - name: Run tests
-        run: sbt 'compile; ++2.13 test'
+      - name: Build core with Maven
+        working-directory: ./stryker4jvm-core
+        run: mvn clean install
```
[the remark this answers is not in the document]
s2176416: When reinstalling locally it is necessary to clean sometimes which is probably why it also exists in the CI but indeed should not have to clean here.

.github/pull_request_template.md
```diff
@@ -1,19 +1,7 @@
-<!---
```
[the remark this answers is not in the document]
s2176416: No idea, will be restored

docs/contributing.md
s2176416: Stryker4k only seems to work with kotlin 1.4.20, this is a limitation in the original project as well

hugo-vrijswijk:
Closing this in favour of #1589. We'll use this PR as the main inspiration, but it is probably easier to go with a cleaner start at this point

PR #1306, opened 2023-02-20
Add back rollback

hugo-vrijswijk:
In #1150, the rollback mechanism was temporarily removed because it was too much extra work with the refactor to fix it. This adds it back


core/src/main/scala/stryker4s/run/RollbackHandler.scala
```diff
@@ -1,17 +1,86 @@
...
-class RollbackHandler {
   def rollbackFiles(
-      // errors: NonEmptyList[CompilerErrMsg],
-      // allFiles: Seq[MutatedFile]
-  ): Either[NonEmptyList[CompilerErrMsg], RollbackResult] = throw new NotImplementedError(
-    "Rollback not implemented yet. If you need rollback, use release version v0.14.3"
-  )
+      errors: NonEmptyList[CompilerErrMsg],
+      allFiles: Vector[MutatedFile]
+  ): Either[NonEmptyList[CompilerErrMsg], RollbackResult] = {
```
hugo-vrijswijk, line 18: This still needs some tests

Issue #1415, opened 2023-09-14
Scala3 Mutations


fabianhjr:
Scala 3 mutations fail to write due to:

```
[error] java.lang.UnsupportedOperationException: Scala212 doesn't support inline modifiers
[error] 	at scala.meta.internal.prettyprinters.TreeSyntax$SyntaxInstances.$anonfun$syntaxTree$1(TreeSyntax.scala:1001)
[..]
[error] 	at scala.meta.prettyprinters.Api$XtensionSyntax.syntax(Api.scala:11)
[error] 	at stryker4s.run.MutantRunner.$anonfun$writeMutatedFile$5(MutantRunner.scala:137)
```

### Stryker4s config

```hocon
stryker4s {
  scala-dialect: "scala3"
}

```

### Stryker4s environment

```
stryker4s 0.14.3+309-78b984b8-SNAPSHOT
```

### Your Environment
| software             | version(s) |
| -------------------- | ---------- |
| Scala version        |   3.3.1    |
| Build tool & version | sbt 1.9.4  | 
| Operating System     | x86_64-linux  |

Issue #1219, opened 2022-09-21
Stryker4s fails on Scala 3 if Union types are used


sergeda:
### Summary
Stryker4s fails on Scala 3 if Union types are used with error:

> java.lang.UnsupportedOperationException: Scala212 doesn't support or types

Here is the [PR](https://github.com/dashkose/zio-http_template/pull/2) were Stryker4s fails 

### Stryker4s config

```hocon
stryker4s {
    timeout: 5000
    thresholds {
        high = 99
        low = 98
        break = 97
    }
    concurrency: 4
    scala-dialect: "scala3"
}
```

### Stryker4s environment
```
stryker4s 0.14.3
```

### Your Environment
| software             | version(s) |
| -------------------- | ---------- |
| Scala version        |     3       |
| Build tool & version |      SBT  1.71    | 
| Operating System     |     Ubuntu 22.04       |

PR #1416, opened 2023-09-14
[core] Use config Dialect to write Scala3 mutations

fabianhjr:
### Fixes #...

Fixes https://github.com/stryker-mutator/stryker4s/issues/1219,
Fixes https://github.com/stryker-mutator/stryker4s/issues/1415


hugo-vrijswijk (approved):
Wow, thanks! This must have been a bug since initial Scala 3 support. I had no idea this was a thing in Scalameta 😅. Thank you so much for the fix!

PR #1424, opened 2023-09-27
[sbt] Add warning in case of empty framework sequence

fabianhjr:
For helping debug an issue on Scala 3.3.1 / sbt 1.9.6

```
[info] set current project to fakeRoot (in build file:/home/fabian/Development/LibreCybernetics/basecodecs.scala/)
[warn] No test frameworks found via sbt / loadedTestFrameworks
[info] Creating 7 test-runners
```

### DIFF
--- a/sbt/src/main/scala/stryker4s/sbt/Stryker4sSbtRunner.scala
+++ b/sbt/src/main/scala/stryker4s/sbt/Stryker4sSbtRunner.scala
@@ -123,6 +123,11 @@ class Stryker4sSbtRunner(
         val javaOpts = extractTaskValue(Test / javaOptions, "javaOptions")
 
         val frameworks = extractTaskValue(Test / loadedTestFrameworks, "test frameworks").values.toSeq
+        if (frameworks.isEmpty)
+          log.warn(
+            "No test frameworks found via loadedTestFrameworks. " +
+              "Will likely result in no tests being run and a NoCoverage result for all mutants."
+          )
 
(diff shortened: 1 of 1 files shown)


sbt/src/main/scala/stryker4s/sbt/Stryker4sSbtRunner.scala
```diff
@@ -123,6 +123,7 @@ class Stryker4sSbtRunner(
         val javaOpts = extractTaskValue(Test / javaOptions, "javaOptions")
 
         val frameworks = extractTaskValue(Test / loadedTestFrameworks, "test frameworks").values.toSeq
+        if (frameworks.isEmpty) log.warn("No test frameworks found via loadedTestFrameworks")
```
hugo-vrijswijk, line 126: I'd like this to be a bit more informative about what it means for the user. Most people probably don't know what `loadedTestFrameworks` is or what impact it has on Stryker4s. Maybe we can add a message that this likely means no tests will run and result in a bogus score? Ideally also some way to fix it, but I don't know what that could be (maybe configuring `testFrameworks` manually?)

fabianhjr: Searched for a while and found the change from sbt-crossproject to projectmatrix, don't know what causes that situation or how to diagnose it further at the moment. UnU

Will change the message as suggested.

hugo-vrijswijk (approved):
Thanks!

PR #1430, opened 2023-10-13
Feature/1363 open report automatically

### LINKED ISSUE #1363: Add option to open HTML report in Stryker4s automatically
Currently the user has to open the report manually, it would be nice to open the report automatically (I could not find if this functionality was present in Stryker4s).

In Stryker.NET this functionality works for both the local report and the dashboard report with the `-o` flag.

YVbakker:
#### What it does

Opens the html report automatically after it's written to disk

Using the new configuration option 'openReportAutomatically'

* Configurable via the sbt cli using `stryker --open-report=true`

closes #1363


hugo-vrijswijk:
Thanks! I think we would need the argument parsing before this could be merged. It might be worth asking how sbt handles argument parsing for tasks like `testOnly` or other things that already parse arguments. 

Ideally, this would also work for other plugins (like maven and command-runner)

modules/core/src/main/scala/stryker4s/config/source/FileConfigSource.scala
```diff
@@ -78,6 +78,8 @@ class FileConfigSource[F[_]](h: ConfigValue[F, Hocon.HoconAt])
   override def testRunnerCommand: ConfigValue[F, String] = read("test-runner.command").as[String]
   override def testRunnerArgs: ConfigValue[F, String] = read("test-runner.args").as[String]
 
+  override def openReport: ConfigValue[F, Boolean] = read("open-report-automatically").as[Boolean]
```
hugo-vrijswijk, line 81: This should probably also be `--open-report`

modules/core/src/main/scala/stryker4s/report/HtmlReporter.scala
```diff
@@ -63,7 +65,16 @@ class HtmlReporter(fileIO: FileIO)(implicit log: Logger) extends Reporter with C
...
-    reportsWriting *>
-      IO(log.info(s"Written HTML report to $indexLocation"))
+    for {
+      _ <- reportsWriting
+      _ <- IO(
+        log.info(s"Written HTML report to $indexLocation")
+      )
+      _ <-
+        if (config.openReport) desktopIO.attemptOpen(indexLocation).handleErrorWith { e =>
+          IO(log.error(s"Error opening report in browser: ${e.getMessage}"))
```
hugo-vrijswijk, line 75:
```suggestion
          IO(log.error("Error opening report in browser", e))
```

modules/core/src/test/scala/stryker4s/report/HtmlReporterTest.scala
```diff
@@ -150,5 +160,39 @@ class HtmlReporterTest extends Stryker4sIOSuite with LogMatchers {
...
+      val fileIOStub = FileIOStub()
+      val desktopIOStub = DesktopIOStub()
+
+      val sut = new HtmlReporter(fileIOStub, desktopIOStub)
+      val report = MutationTestResult(thresholds = Thresholds(100, 0), files = Map.empty)
+      val metrics = Metrics.calculateMetrics(report)
+
+      sut
+        .onRunFinished(FinishedRunEvent(report, metrics, 10.seconds, fileLocation)) >>
+        desktopIOStub.openCalls.asserting { calls =>
```
hugo-vrijswijk, line 193:
```suggestion
        desktopIOStub.openCalls.assertEquals(Seq.empty)
```

hugo-vrijswijk (approved):
LGTM!! 🚀

PR #1589, opened 2024-06-19
Add language mutator-api submodule

hugo-vrijswijk:
This adds a new submodule called `mutator-api` which contains interfaces that language-specific mutators can implement.

Interfaces are to be added and discussed before merging this PR.

@jessemaas  @YVbakker

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -30,7 +30,8 @@ lazy val root = (project withId "stryker4s" in file("."))
       sbtTestRunner.projectRefs ++
       testRunnerApi.projectRefs ++
       api.projectRefs ++
-      testkit.projectRefs) *
+      testkit.projectRefs :+
+      mutatorApi.project) *
   )
 
 lazy val core = (projectMatrix in file("modules") / "core")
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -64,6 +65,12 @@ lazy val api = (projectMatrix in file("modules") / "api")
   .settings(commonSettings, apiSettings)
   .jvmPlatform(scalaVersions = versions.fullCrossScalaVersions)
 
+/** Java-only project that contains interfaces for a language-agnostic mutator API
+  */
+lazy val mutatorApi = (project in file("modules") / "mutatorApi")
+  .settings(mutatorApiSettings)
+// .dependsOn(testkit.jvm(versions.scala3) % Test)
```
hugo-vrijswijk, line 72: I've added this and commented _if_ we want to add any tests. We could also add junit and write them in Java, but this seemed less work. I think for now we don't need any tests

hugo-vrijswijk:
This PR has become a bit stale. Can be reopened if interest for this comes back

PR #1701, opened 2025-02-13
docs: update github link for full list of scala dialect

moronyoh:
#### What it does
Fix one broke link in the documentation


hugo-vrijswijk (approved):
Thanks!

PR #1755, opened 2025-04-22
fix: no-coverage mutants showing reported as static

### LINKED ISSUE #1754: Stryker4s incorrectly calculates mutant score for uncovered code
According to the documentation, Stryker is supposed to compute two metrics : mutant coverage, and mutant coverage on coverage code.
By default, mutant coverage appears to be the standard, though with other frameworks both are at least displayed in the dashboard while Stryker only outputs a single metric.

Within the `MutantRunner` class, the code for `noCoverageMutant` appears to be directly and erroneously copied from `staticMutant`, which ultimately leads to all code that isn't covered by a test being considered "ignored" rather than "no coverage".
This then causes mutationScore and mutationScoreBasedOnCoveredCode in `MetricsResult` to be identical.

This is simple to reproduce with a sample project with one class with two methods, one tested and one not, and the score will be 100% if the one method is tested thoroughly.

andrewl102:
#### What it does

Correctly calculate the mutation score on ignored code.

#### How it works

Fix copy and pasted method used for calculating no coverage code.

### DIFF
--- a/modules/core/src/main/scala/stryker4s/run/MutantRunner.scala
+++ b/modules/core/src/main/scala/stryker4s/run/MutantRunner.scala
@@ -238,10 +238,9 @@ class MutantRunner(
 
   private def noCoverageMutant(mutant: MutantWithId): MutantResult = mutant
     .toMutantResult(
-      MutantStatus.Ignored,
-      statusReason =
-        "This is a 'static' mutant and can not be tested. If you still want to have this mutant tested, change your code to make this value initialize each time it is called.".some
+      if (config.noCoverageIgnored) MutantStatus.NoCoverage else MutantStatus.Ignored,
+      statusReason = "This is a no coverage mutant and is not tested.".some
     )
-    .copy(static = true.some)
+    .copy(static = None)
 
 }
--- a/modules/core/src/main/scala/stryker4s/run/MutantRunner.scala
+++ b/modules/core/src/main/scala/stryker4s/run/MutantRunner.scala
@@ -238,9 +238,8 @@ class MutantRunner(
 
   private def noCoverageMutant(mutant: MutantWithId): MutantResult = mutant
     .toMutantResult(
-      if (config.noCoverageIgnored) MutantStatus.NoCoverage else MutantStatus.Ignored,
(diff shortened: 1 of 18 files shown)


modules/core/src/main/scala/stryker4s/run/MutantRunner.scala
```diff
@@ -238,10 +238,9 @@ class MutantRunner(
...
   private def noCoverageMutant(mutant: MutantWithId): MutantResult = mutant
     .toMutantResult(
-      MutantStatus.Ignored,
-      statusReason =
-        "This is a 'static' mutant and can not be tested. If you still want to have this mutant tested, change your code to make this value initialize each time it is called.".some
+      if (config.noCoverageIgnored) MutantStatus.NoCoverage else MutantStatus.Ignored,
+      statusReason = "This is a no coverage mutant and is not tested.".some
     )
-    .copy(static = true.some)
+    .copy(static = None)
```
hugo-vrijswijk, line 244: You can remove the `.copy` as the default value for static is already `None`.

hugo-vrijswijk (commented):
Thanks for this! I have seen mutants show up as static that I didn't think should be, and always wondered why 😅. So I am very happy you found this. IMO this is a bugfix and the old behaviour is not desired. So you can remove the configuration options. Unfortunate as that is most of the lines changed. Sorry 🙈.

The [mutant metrics](https://stryker-mutator.io/docs/mutation-testing-elements/mutant-states-and-metrics/#metrics) we have defined have a `mutationScoreBasedOnCoveredCode`, as you noticed. But it is not used in Stryker4s anywhere. Both values can be interesting, but not something that should be behind a configuration option, I think. My suggestion would be to add the `mutationScoreBasedOnCoveredCode` value to the `ConsoleReporter` instead. Maybe something similar to how the [html report](https://dashboard.stryker-mutator.io/reports/github.com/stryker-mutator/stryker4s/master#mutant) also shows total and covered scores. 

This is kind of 2 different things: a bugfix where no-coverage is reported as static, and reporting the score based on covered code. Ideally they would be separate PR's, but I can separate the changelog generation for these if it stays the same PR.

PR #1871, opened 2025-10-30
build: pin third-party actions to a full-length commit SHA

### LINKED ISSUE #1870: Third-party GitHub Actions are not pinned to full-length commit SHAs
Some third-party actions in this repository are currently referenced by version tags (e.g., @v1) instead of being pinned to a full-length commit SHA.

This poses a potential security risk since tags can be moved or modified by repository owners, allowing malicious code to be injected into workflows without any visible change in the workflow file.

**Impact**:

- A compromised or modified action could access repository secrets and the GITHUB_TOKEN.
- This could lead to unauthorized modifications or data exfiltration.

https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

jaspervdveen:
#### What it does

This PR updates all third-party GitHub Actions used in the repository workflows to be pinned to full-length commit SHAs instead of floating version tags (e.g., `@v1`).

And I replaced a deprecated action with an action from GitHub.

This change aligns with GitHub’s security best practices 
https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

#### How it works

- Replaced version tags for third-party actions with their corresponding full commit SHAs.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -19,7 +19,7 @@ jobs:
         with:
           app-id: ${{ vars.STRYKER_MUTATOR_APP_ID }}
           private-key: ${{ secrets.STRYKER_MUTATOR_NPA_KEY }}
-      - uses: googleapis/release-please-action@16a9c90856f42705d54a6fda1823352bdc62cf38
+      - uses: googleapis/release-please-action@16a9c90856f42705d54a6fda1823352bdc62cf38 # v4.4.0
         with:
           # Version is determined by sbt-ci-release (which uses sbt-dynver)
           # Actual release code is in `./ci.yml`


jaspervdveen:
@nicojs @hugo-vrijswijk GitHub offers policies at the repository and organization level to require actions to be pinned to a full-length commit SHA. I think it's a good idea to enable this policy to prevent supply chain attacks.

hugo-vrijswijk (commented):
@jaspervdveen  I'm fine with enabling this org-wide but do have some concerns/questions. Does dependabot/renovatebot generate update PR's for these? And if there are updates, are they made for each commit, or only for hashes that are linked to a tag?

If not, this really decreases the maintainability of using actions, as it becomes much more difficult to know when there's breaking change, and stay up to date. Which in itself is a security concern

jaspervdveen:
> Does dependabot/renovatebot generate update PR's for these? And if there are updates, are they made for each commit, or only for hashes that are linked to a tag?

Yes, Dependabot and Renovate should support generating update PR's for new tagged versions, as long as you add the version tag comment after the SHA.

Renovate: https://docs.renovatebot.com/modules/manager/github-actions/#digest-pinning-and-updating
Dependabot: I can't find any clear documentation about it, other than [this](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot). 
However, I saw the following example at a conference this week:
image
Reference: https://github.com/joshjohanning/fortifying-workflows-slides/blob/main/TNL25_Josh%20Johanning_Fortifying%20Your%20Workflows.pdf

hugo-vrijswijk (approved):
Looks good to me then! :shipit:

Issue #1417, opened 2023-09-15
Improve error message about unrealistic result


hugo-vrijswijk:
If Stryker4s exits too fast, all mutations are survived/killed or maybe some other condition happens, there could be more informative logging.

Example of logging now:

```

Issue #646, opened 2020-11-16
Stryker4s always report score of 0% with sbt-crossproject


hugo-vrijswijk:
### Summary

When using [sbt-crossproject](https://github.com/portable-scala/sbt-crossproject), Stryker4s will always report a mutation score of 0%. Stryker4s will still mutate, write new source-code and run mutation tests, but it appears the classpath includes only the JVM-only sources, instead of all sources.

### Stryker4s environment

```
stryker4s 0.10.0-RC3
```

### Your Environment
| software             | version(s) |
| -------------------- | ---------- |
| Scala version        |      Any      |
| Build tool & version |    sbt-scalajs-crossproject 1.0.0        | 
| Operating System     | Windows (WSL)           |

PR #1994, opened 2026-05-16
fix(sbt): scope task queries to the invoking project

rwaldvogel:
### Related: #646, #1417, #1424

Same silent-0% failure mode previously reported in #646 (sbt-crossproject) and #1417 (UX of the bad result). #1424 added the *warning* that fires when this happens; this PR addresses the underlying cause.

#### What it does

Makes `sbt sub/stryker` actually work when the project being mutated lives in a subdirectory (`lazy val app = (project in file("app"))`) rather than at the build root. With the current 0.20.x plugin, that layout silently produces a 0% mutation score: the mutator runs, mutants are generated, "Initial test run succeeded" prints — but every mutant survives because no tests actually execute. Reproduces on a 12-line bare repro on Scala 3 and Scala 2.13; only difference vs. the working case is `(project in file("subdir"))` vs. an implicit root.

Also fixes a second, smaller issue uncovered while validating the first: `blocklistedScalacOptions` only strips `-Xfatal-warnings`, not the newer `-Werror`. Any project that opts into strict warnings (default for tpolecat 0.5.3+ on Scala 3) hits a compile failure during mutation switching whose root cause is a normal warning (e.g. non-exhaustive `match` over the synthetic `Some(\"1\") | Some(\"2\") | _` cases) being promoted to an error.

#### How it works

**Issue 1 — task scope.** `Stryker4sSbtRunner.setupSbtTestRunner` queried `Test / loadedTestFrameworks`, `Test / fullClasspath`, `Compile / compile`, `Test / testGrouping` (and `Test / executeTests` in the legacy runner) without a project scope. `PluginCompat.runTask` then resolved them against the State's current ref — which, for \`sub/stryker\`, is the auto-generated root, not the subproject. The root has no test frameworks and an empty test classpath, so the empty-framework warning added in #1424 fired and the run silently produced 0%.

The settings appended via \`extracted.appendWithSession\` (scalacOptions, Test/fork, Compile/unmanagedSourceDirectories, Test/javaOptions, Test/testOptions, libraryDependencies, resolvers, logManager, and the ACTIVE_MUTATION javaOption) had the same problem at a quieter scope — they applied build-globally rather than to the subproject.

Fix: capture \`thisProjectRef\` at the task macro site in \`Stryker4sPlugin.strykerTask\` (where it correctly resolves to the calling subproject) and thread it through \`Stryker4sSbtRunner\` and \`LegacySbtTestRunner\`. Every task query and every appended setting is now \`targetProject / ...\`-scoped.

**Issue 2 — \`-Werror\`.** One-line addition to the existing strip list.

#### Test plan

- New scripted test \`test-subproject\` defines a project in \`file(\"app\")\`, Scala 3 LTS (3.3.7, binary-compatible with the 3.8.x where the issue first surfaced), \`scalacOptions ++= Seq(\"-Werror\", \"-Xfatal-warnings\")\`, and a 50% break threshold. Invokes \`> app/stryker\`. If either fix regresses (scope or \`-Werror\` strip), the score collapses to 0% and the scripted test fails.
- CI now publishes \`sbtTestRunner3\` alongside \`sbtTestRunner2_12\` so the Scala 3 test classpath resolves.
- Verified locally on both \`sbtPlugin/scripted\` (sbt 1 / Scala 2.12 plugin) and \`sbtPlugin3/scripted\` (sbt 2 / Scala 3 plugin): test-subproject 100% (3/3 killed); existing test-1 unchanged at 88.89%.

No new dependencies, no API changes beyond an internal constructor parameter on two non-public runner classes, no behaviour change for builds already using a single root project.

### DIFF
--- a/modules/sbt/src/main/scala/stryker4s/sbt/Stryker4sPlugin.scala
+++ b/modules/sbt/src/main/scala/stryker4s/sbt/Stryker4sPlugin.scala
@@ -170,10 +170,6 @@ object Stryker4sPlugin extends AutoPlugin {
     }
 
     val sbtConfig = SbtConfigSource[IO]().value
-    // The project the user invoked `stryker` on. Captured here (task-scoped) so that the runner
-    // can scope its task queries (Test / loadedTestFrameworks, Compile / compile, ...) to it.
-    // Without this, `sbt sub/stryker` queries those tasks against the build's current project
-    // (typically the auto-generated root), which has no test frameworks → silent 0% mutation score.
     val targetProject = thisProjectRef.value
 
     Def.task {
--- a/modules/sbt/src/main/scala/stryker4s/sbt/Stryker4sPlugin.scala
+++ b/modules/sbt/src/main/scala/stryker4s/sbt/Stryker4sPlugin.scala
@@ -170,6 +170,11 @@ object Stryker4sPlugin extends AutoPlugin {
     }
 
     val sbtConfig = SbtConfigSource[IO]().value
+    // The project the user invoked `stryker` on. Captured here (task-scoped) so that the runner
+    // can scope its task queries (Test / loadedTestFrameworks, Compile / compile, ...) to it.
+    // Without this, `sbt sub/stryker` queries those tasks against the build's current project
+    // (typically the auto-generated root), which has no test frameworks → silent 0% mutation score.
+    val targetProject = thisProjectRef.value
 
     Def.task {
       implicit val runtime: IORuntime = IORuntime.global
@@ -180,7 +185,7 @@ object Stryker4sPlugin extends AutoPlugin {
       val extraConfigSources = List(sbtConfig, cliConfig)
 
       Deferred[IO, FiniteDuration] // Create shared timeout between testrunners
-        .map(new Stryker4sSbtRunner(state.value, javaHome.value, _, extraConfigSources))
+        .map(new Stryker4sSbtRunner(state.value, javaHome.value, targetProject, _, extraConfigSources))
         .flatMap(_.run())
         .flatMap {
           case ErrorStatus => IO.raiseError(new MessageOnlyException("Mutation score is below configured threshold"))
--- a/modules/sbt/src/main/scala/stryker4s/sbt/Stryker4sSbtRunner.scala
+++ b/modules/sbt/src/main/scala/stryker4s/sbt/Stryker4sSbtRunner.scala
@@ -33,6 +33,7 @@ import Stryker4sPlugin.autoImport.stryker
 class Stryker4sSbtRunner(
     state: State,
     javaHome: Option[File],
+    targetProject: ProjectRef,
     sharedTimeout: Deferred[IO, FiniteDuration],
     override val extraConfigSources: List[ConfigSource[IO]]
 )(implicit
@@ -53,14 +54,14 @@ class Stryker4sSbtRunner(
         LogManager.defaultManager(ConsoleOut.printStreamOut(new PrintStream((_: Int) => {})))
 
       val fullSettings = settings ++ Seq(
-        logManager := {
+        targetProject / logManager := {
           if ((stryker / logLevel).value == Level.Debug) logManager.value
           else emptyLogManager
         }
       )
       val newState = extracted.appendWithSession(fullSettings, state)
 
-      NonEmptyList.of(Resource.pure(new LegacySbtTestRunner(newState, fullSettings, extracted)))
+      NonEmptyList.of(Resource.pure(new LegacySbtTestRunner(newState, fullSettings, extracted, targetProject)))
     }
 
     def setupSbtTestRunner(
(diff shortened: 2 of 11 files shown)


modules/sbt/src/sbt-test/sbt-stryker4s/test-subproject/app/stryker4s.conf
```diff
@@ -0,0 +1,3 @@
+stryker4s {
```
hugo-vrijswijk: You can remove this file as the relevant configuration is already done in the build.sbt, and `console` is in the list of default reporters

modules/sbt/src/sbt-test/sbt-stryker4s/test-subproject/README.md
```diff
@@ -0,0 +1,14 @@
+# SBT plugin test project — subproject layout
```
hugo-vrijswijk, line 1: I think this readme is plenty documentation. In most other places where the lines of code changed are less than the documentation it reads too verbose. This file explains the case fine, so should be the main point, where in other cases it is fairly clear what is happening

modules/sbt/src/main/scala/stryker4s/sbt/Stryker4sPlugin.scala
```diff
@@ -170,6 +170,11 @@ object Stryker4sPlugin extends AutoPlugin {
     }
 
     val sbtConfig = SbtConfigSource[IO]().value
+    // The project the user invoked `stryker` on. Captured here (task-scoped) so that the runner
+    // can scope its task queries (Test / loadedTestFrameworks, Compile / compile, ...) to it.
+    // Without this, `sbt sub/stryker` queries those tasks against the build's current project
+    // (typically the auto-generated root), which has no test frameworks → silent 0% mutation score.
+    val targetProject = thisProjectRef.value
```
hugo-vrijswijk, line 173: The comment here is again a bit too verbose. My suggestion would be to either remove the comment altogether or change it to a short line explaining why. It is already quite obvious that we _should_ be targeting the current project to run stryker on

modules/sbt/src/main/scala/stryker4s/sbt/Stryker4sSbtRunner.scala
```diff
@@ -171,8 +175,12 @@ class Stryker4sSbtRunner(
...
         // -Ywarn for Scala 2.12, -W for Scala 2.13
       ).flatMap(opt => Seq(s"-Ywarn-$opt", s"-W$opt")) ++ Seq(
-        // Disable fatal warnings, as they will cause a lot of mutation switching statements to not compile
-        "-Xfatal-warnings"
+        // Disable fatal warnings, as they will cause a lot of mutation switching statements to not compile.
+        // -Xfatal-warnings is the Scala 2.x flag; -Werror is the Scala 3 / tpolecat 0.5.3+ replacement
+        // and must also be stripped — otherwise warnings the mutation-switching pattern triggers
+        // (e.g. non-exhaustive match on the synthetic Some("n") cases) propagate as errors.
+        "-Xfatal-warnings",
+        "-Werror"
```
hugo-vrijswijk: This is a great fix, but should probably be in a separate PR

modules/sbt/src/main/scala/stryker4s/sbt/Stryker4sSbtRunner.scala
```diff
@@ -185,15 +193,18 @@ class Stryker4sSbtRunner(
       }
       log.debug(s"System properties added to the forked JVM: ${filteredSystemProperties.mkString(",")}")
 
+      // All settings are scoped to targetProject so that appendWithSession changes only affect the
+      // subproject the user invoked stryker on, not every project in the build (in particular not the
+      // auto-generated root, against which our task queries previously resolved).
```
hugo-vrijswijk: Here the comments are again a bit too verbose for me. You don't need to explain how this previously worked

modules/sbt/src/sbt-test/sbt-stryker4s/test-subproject/build.sbt
```diff
@@ -0,0 +1,27 @@
+// The whole point of this scripted test: the project to be mutated is in `file("app")`,
+// not at the build root. With the pre-fix plugin, `sbt app/stryker` resolved task queries
+// against the auto-generated root (no test framework) → 0% mutation score.
+
+// Scala 3 LTS so we can exercise the -Werror stripping path: -Werror is what
+// tpolecat 0.5.3+ emits for Scala 3 (replacing the legacy -Xfatal-warnings).
+// This is also the path the original bug surfaced on (Scala 3.8.3 in a real-world
+// build, binary-compatible with 3.3.x).
```
hugo-vrijswijk: You can probably remove every single comment in this file. It is already clear why most statements are added. And if not, they are explained in the readme.

hugo-vrijswijk (changes requested):
Thanks for this PR! Was (part of) this created using AI? Most of the fix seems good.
It should be noted that some of the issues linked (like #646 and #1417) are not actually related, although #1417 somewhat is.

Running Stryker4s on multi-module projects was also always possible, but required the syntax of `sbt 'project core; stryker'`. Which, admittedly, is not as nice. So I am glad that with this we can use the more familiar syntax of `sbt 'core/stryker'`!

As for the PR itself, I've left a few comments. A lot of the code comments are quite verbose and unnecessary. I've left a few comments that should be easy enough to fix.

rwaldvogel:
Thanks for taking the time to review my PR.

Let me address the elephant in the room first: yes, I use(d) AI for the changes leading to this PR. Working solo on this kind of thing, I lean on Claude Code with Opus 4.7 in a driver/navigator style — me the navigator.

I appreciate the way you handle comments — it's my personal preference too, code has to stand by itself without comments. Happy to remove them.

On the linked issues: you're right — #646 is sbt-crossproject specific and not addressed here, I'll drop it. #1417 stays as "somewhat related" since this fix makes the silent-0% case less common but doesn't add the diagnostics that issue actually asks for. #1424 stays as direct.

PR splitting: will do. Plan is to land this one first (comments stripped, description tightened), then the `-Werror` fix as a follow-up — let me know if you'd prefer the other order.

hugo-vrijswijk (approved):
Thanks for the quick changes! This is good to merge

rwaldvogel:
Hi Hugo,

just a quick thank-you for the fast turnarounds on #1994 and #1999 — the careful review and the quick release out the door really helped me unblock our downstream project. Much appreciated!

Issue #321, opened 2019-12-05
Stryker4s error when using scalacOption "-Ywarn-unused:patvars"


hugo-vrijswijk:
### Summary

When having the scalacOption `"-Ywarn-unused:patvars"` (warn on unused pattern match variables) enabled in combination with `"-Xfatal-warnings"` (fail compilation on warnings), Stryker4s is very likely to fail.

This is because of the built pattern matches for mutation switching. The chances of a variable being unused in one of the pattern matches is very high.

The solution here is probably to remove the scalacOption before running Stryker. In sbt, this is not too difficult, but we should make sure to return the **original** state when finishing stryker, so no options are changed for subsequent commands. For maven, a bit more investigation is needed.

PR #1999, opened 2026-05-20
fix(sbt): strip -Werror alongside -Xfatal-warnings

rwaldvogel:
Scala 3 / tpolecat 0.5.3+ counterpart of `-Xfatal-warnings` (#321) — same failure mode, same fix.


hugo-vrijswijk (approved):
TY!

